PR #9, opened 2016-12-29
Add function to build template values

tomwilkie:


grafanalib/core.py
```diff
@@ -213,6 +213,31 @@ def DashboardLink(dashboard, uri, keepTime=True, title=None):
     }
 
 
+def Template(default, dataSource, label, name, query):
```
jml: Took me a while to figure out what 'default' means. Can you think of a better name? No worries if not.

jml: Actually, after looking at https://github.com/weaveworks/monitoring/pull/113 I have no idea what it means.

tomwilkie: AFAICT is the "default" value for this variable - ie, when you visit the page for the first time.

PR #10, opened 2017-01-04
Support Alert configuration in graph for Grafana 4.x

astronaut1712:


grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+
+
+def Alert(alertConditions=[],
+          executionErrorState="alerting",
+          frequency="60s",
+          handler=1,
+          noDataState="no_data",
+          message="Alert message",
+          name="Alert name",
+          notifications=[]):
```
jml: See above comment re mutable default arguments.

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+    }
+
+
+def Alert(alertConditions=[],
+          executionErrorState="alerting",
+          frequency="60s",
+          handler=1,
+          noDataState="no_data",
+          message="Alert message",
+          name="Alert name",
```
jml: We should only provide defaults if they don't need to be changed in most cases. An alert that tells you "Alert name: alert message" definitely needs to be changed.

Could you please delete the defaults for `name` and `message` here?

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
 )
 
 
+def AlertCondition(target, queryParams=["A", "5m", "now"],
+                   evaluator=["gt", 5], operator="and", type="avg"):
+    """
+    Alert conditions:
+        :param target: Target
+        :param queryParams: [refId, in_time, "now"]
+        :param evaluator: [gt/lt/within_range/outside_range/no_value, values]
```
jml: Can you please define constants for these (similar to the ones we use for axes, e.g. `SHORT_FORMAT`), and use those in the docstring and the function body?

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+        :param operator: and/or
```
jml: Similarly, please provide constants for these.

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
 )
 
 
+def AlertCondition(target, queryParams=["A", "5m", "now"],
+                   evaluator=["gt", 5], operator="and", type="avg"):
```
jml: Please rename the `type` parameter to `reducerType`.

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+    """
+    Alert conditions:
+        :param target: Target
+        :param queryParams: [refId, in_time, "now"]
+        :param evaluator: [gt/lt/within_range/outside_range/no_value, values]
+        :param operator: and/or
+        :param type: avg/min/max/sum/count/last/median
```
jml: And these 😄

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+        },
+        "reducer": {
+            "params": [],
+            "type": type
+        },
+        "type": "query"
+    }
+
+
+def Alert(alertConditions=[],
```
jml: It's generally [a bad idea to have mutable default arguments](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments). Can you please change this so the list arguments default to `None`, with the actual default being assigned inside the body of the function?

grafanalib/core.py
jml: Both `evaluator` and `queryParams` look complex enough that they should probably have their own top-level functions, so that we say something like:

```python
AlertCondition(
    target=Target(...), 
    evaluator=GreaterThan(5),
    ...
)
```

I don't really understand what `queryParams` is doing well enough to make a similarly concrete suggestion—what _is_ it for?

grafanalib/core.py
```diff
@@ -299,6 +302,61 @@ def TimePicker(refreshIntervals, timeOptions):
...
+        },
+        "query": {
+            "model": target,
+            "params": queryParams
+        },
+        "reducer": {
+            "params": [],
+            "type": type
+        },
+        "type": "query"
```
jml: Trailing comma please, and also probably make `type` a parameter that defaults to `"query"`.

astronaut1712:
Thanks @jml. That's good suggestions, I updated the source code.

Regards,

jml:
Thanks!

I've done another round of review and applied it directly in #11 to save an extra roundtrip.

PR #12, opened 2017-01-24
Add quay.io prefix to gfdatasource image

jpellizzari:
@tomwilkie This solved an issue for me when standing up a `minikube` k8s cluster. Also, can I pretty please have write access to the repo via our org?


jml:
> Also, can I pretty please have write access to the repo via our org?

Oops, my fault. You & everyone else at Weave should have write permissions now.

PR #13, opened 2017-01-24
Stop pushing to Docker Hub

jml:
We're getting 404s when pushing & pulling from
hub.docker.io/weaveworks/gfdatasource. Just stop pushing/pulling there and use
only quay.io.


tomwilkie:
LGTM

PR #14, opened 2017-01-27
Stack Weave QPS graphs

jml:
We talked about this a little bit in the London office.

Justification is that when traffic switches from 200s to 500s, it's hard to tell how much overall traffic is being affected.


tomwilkie:
LGTM

PR #16, opened 2017-02-02
Add make deps command

paulbellamy:
Works for me, but no idea if it is the correct thing to do for python...


setup.py
```diff
@@ -35,6 +35,11 @@ def local_file(name):
     ],
     install_requires=[
     ],
+    extras_require={
+        'dev': [
+            'flake8',
+        ],
+    },
```
jml: This particular change isn't strictly necessary for the Makefile stuff, but it's a bit more idiomatic. requirements.txt files are more often used for pinning versions in an application than for library dependencies.

.gitignore
```diff
@@ -4,3 +4,4 @@ __pycache__
 build/
 dist/
 .uptodate
+/.env
```
paulbellamy: does `/.env` mean what you think?

circle.yml
```diff
@@ -9,8 +9,7 @@ machine:
 
 dependencies:
   override:
-    - pip install -e .
-    - pip install -r dev-requirements.txt
+    - pip install -e .[dev]
```
paulbellamy: no `make deps` here for consistency?

jml: Good point! Done.

.gitignore
jml: Yep.

PR #17, opened 2017-03-02
Objects not functions

jml:
I started this work a couple of weeks ago before I went on leave, and then dug it up on my tube journey in. 

No behavioural changes here, but sets us up so that we can get rid of the `next(GRAPH_ID)` nonsense. Also adds tests, which are pretty important for Python projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build/
 dist/
 .uptodate
 /.env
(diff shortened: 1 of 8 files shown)


jml:
Merging without review, because begging for reviews on our internal slack is a waste of time.

PR #22, opened 2017-03-27
SingleStat panel class added

dmajere:
Here is a singlestat class

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -22,6 +22,16 @@ class RGBA(object):
 
 
 @attr.s
+class RGB(object):
+    r = attr.ib(validator=instance_of(int))
+    g = attr.ib(validator=instance_of(int))
+    b = attr.ib(validator=instance_of(int))
+
+    def to_json_data(self):
+        return "rgb({}, {}, {})".format(self.r, self.g, self.b)
+
+
+@attr.s
 class Pixels(object):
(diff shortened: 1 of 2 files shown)


grafanalib/core.py
```diff
@@ -646,3 +694,151 @@ def to_json_data(self):
...
+            'maxValue': self.maxValue,
+            'minValue': self.minValue,
+            'show': self.show,
+            'thresholdLabels': self.thresholdLabels,
+            'thresholdMarkers': self.thresholdMarkers,
+        }
+
+
+@attr.s
+class SingleStat(object):
```
jml: Could you please add a docstring for this?

jml:
Lovely, thanks. Could you please add a changelog entry?

grafanalib/core.py
dmajere: Done

dmajere:
what version it will be? 0.1.3?

jml:
> what version it will be? 0.1.3?

Just call it `NEXT` and we'll roll dice the next time we do a release  :)

jml:
Actually, I just pushed a section for this in 1c72a4dd80da001c20e7deb471db352dce7b5317

dmajere:
Done

dmajere:
I'll do the same for other two

PR #28, opened 2017-03-28
XAxis class update: additional fields

dmajere:
Xaxis class update.
`mode`, `name`, `values` fields added.


jml:
Thanks again! I'll merge after you add a changelog entry.

dmajere:
Changelog record added

Issue #26, opened 2017-03-28
Small improvement proposal on Legend class


dmajere:
There are four class fields that define what type of values grafana will show in legend

```
    avg = attr.ib(default=False, validator=instance_of(bool))
    current = attr.ib(default=False, validator=instance_of(bool))
    max = attr.ib(default=False, validator=instance_of(bool))
    min = attr.ib(default=False, validator=instance_of(bool))
```

But, grafana won't actually show them if you wont set `values` field to True too.
May be we should automatically set `values` to True if any of `avg`, `current`, `max` or `min` are set to True

jml:
SGTM. `values` should default to `None`, and if it is `None`, we'll set its value to `any([avg, current, max, min])`. If it's set explicitly, we should respect that explicit setting.

dmajere:
Closed by #30

PR #30, opened 2017-03-29
Issue26: Legend class field  should be set based on other values

dmajere:
Related to #26 
I wanted to do this with `attr.Attribute.convert`, but attr passes only field value to this callable, so there is no way to inspect other field values. 
On the other hand `attr.Attribute.validate` gets `instance`, `attribute`, and `value`. But setting field value in validation function is obscure.
So, I've just made an explicit method.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -172,7 +172,7 @@ class Legend(object):
     min = attr.ib(default=False, validator=instance_of(bool))
     show = attr.ib(default=True, validator=instance_of(bool))
     total = attr.ib(default=False, validator=instance_of(bool))
-    values = attr.ib(default=False, validator=instance_of(bool))
+    values = attr.ib(default=None)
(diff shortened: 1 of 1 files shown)


grafanalib/core.py
```diff
@@ -172,14 +172,20 @@ class Legend(object):
...
     rightSide = attr.ib(default=False, validator=instance_of(bool))
     sideWidth = attr.ib(default=None)
 
+    def convert_values_field_value(self):
+        if self.values is None:
+            self.values = (self.avg or self.current or self.max or self.min)
+
     def to_json_data(self):
+        self.convert_values_field_value()
+
```
jml: Sorry to be fussy about this, but I'd much rather something that doesn't change state.

e.g.

```python
values = (self.avg or self.current or self.max or self.min) if self.values is None else self.values
```

dmajere:
Redone according to comment

PR #34, opened 2017-04-05
Custom validators moved to dedicated file. Tests for custom validator.

dmajere:
Hi, I want to move all custom validators into dedicated file


jml:
Good call!

PR #47, opened 2017-07-10
Add app initialisation options to gfdatasource

leth:
- [x] Update README
- [x] Fix `__`
- [x] Fix sorting of args
- [x] Remove disable flag
- [x] Check subparsers

### DIFF
--- a/gfdatasource/gfdatasource
+++ b/gfdatasource/gfdatasource
@@ -1,13 +1,15 @@
 #!/usr/bin/env python3
 
-"""Tools for maintaining Grafana datasources."""
+"""Tools for maintaining Grafana datasources & application plugins."""
 
 import argparse
-import attr
-import requests
+import json
 import sys
 import time
 from urllib.parse import ParseResult, urlparse
+
+import attr
+import requests
 
 
 @attr.s
@@ -46,6 +48,33 @@ class DataSource(object):
 
 
 @attr.s
+class App(object):
+    '''An App plugin for Grafana
+
+    Grafana application configuration logic is entirely client side, therefore
+    this util cannot fill UI fields without emulating a full browser.
+    Therefore this must be provided with the right data to be sent to the
+    server to emulate the output of the client-side application.
+    '''
+
+    id = attr.ib()
+    enabled = attr.ib()
+    json_data = attr.ib()
+    secure_json_data = attr.ib()
+
+    def to_json_dict(self):
+        data = {
+            'id': self.id,
+            'enabled': self.enabled,
+            'pinned': self.enabled,
+            'json_data': self.json_data,
+        }
+        if self.secure_json_data is not None:
+            data['secure_json_data'] = self.secure_json_data
+        return data
+
+
+@attr.s
 class GrafanaAPI(object):
     """HTTP client for Grafana's API."""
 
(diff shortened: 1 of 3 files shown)


jml:
This is kind of hard to review because of the rename. What's the purpose of this change? How will people take advantage of it? (I guess README will answer that). Are the command-line args of current version compatible with this one?

leth:
> This is kind of hard to review because of the rename.

I'll try pulling that out to a separate commit, if not will restore the name.

> What's the purpose of this change?

Grafana has 'plugins' and 'apps'; this allows you to enable an 'app', which in turn enables bundled plugins, which will allow us to add a new data source.

> How will people take advantage of it? (I guess README will answer that). 

Yeah will get to that!

> Are the command-line args of current version compatible with this one?

Should be, yes

gfdatasource/gfdatasource
```diff
@@ -103,19 +235,30 @@ def _split_creds(url):
...
+        cmd_func = _cmds[opts.cmd]
+    except KeyError:
+        print('Unknown command', opts.cmd)
+        sys.exit(1)
+
+    update_func = cmd_func(grafana_api, opts)
+
     while True:
-        grafana_api.update_datasource(datasource)
+        update_func()
```
jml: Why set `update_func` only to call it here?

gfdatasource/gfdatasource
```diff
@@ -59,33 +88,136 @@ class GrafanaAPI(object):
...
-        help="How frequently to update Grafana, in seconds",
+
+    app_parser = subparsers.add_parser('app')
+    app_parser.add_argument(
+        '--id', type=str, required=True, help="The app plugin ID",
+    )
+    app_parser.add_argument(
+        '--disabled', type=bool, default=False,
+        help="Whether the app should be disabled",
+    )
```
jml: Why would you want to use this?

gfdatasource/gfdatasource
```diff
@@ -59,33 +88,136 @@ class GrafanaAPI(object):
...
+    __default_subparser = None
+
+    def set_default_subparser(self, name):
+        self.__default_subparser = name
+
+    def parse_args(self, args=None, namespace=None):
+        subparser_action = None
+        for action in self._subparsers._actions:
+            if isinstance(action, argparse._SubParsersAction):
+                subparser_action = action
```
jml: Why do we want to set this to the last `_SubParsersAction` in `_actions`?

gfdatasource/gfdatasource
```diff
@@ -59,33 +88,136 @@ class GrafanaAPI(object):
...
+    )
+
+    def update_app():
+        grafana_api.update_app(app)
+
+    return update_app
+
+
+class DefaultSubcommandArgParse(argparse.ArgumentParser):
+    __default_subparser = None
```
jml: Personally, I avoid `__foo` variables, preferring `_foo` instead, as the name-mangling only helps in tiny codebases. However, feel free to leave as is.

gfdatasource/gfdatasource
```diff
@@ -59,33 +88,136 @@ class GrafanaAPI(object):
...
+        secure_json_data=secure_json_data,
+    )
+
+    def update_app():
+        grafana_api.update_app(app)
+
+    return update_app
+
+
+class DefaultSubcommandArgParse(argparse.ArgumentParser):
```
jml: I don't really understand what this is for.

gfdatasource/gfdatasource
leth: Good spot; I copied the original version from the interwebs.

gfdatasource/gfdatasource
leth: I have no idea! I'll remove it.

gfdatasource/gfdatasource
leth: Previously the command parser understood `<cmd> <*options>`
Now it understands `<cmd> <global opts> [datasource|app] <specific opts>`
This is here to make `<cmd> <global opts> <datasource opts>` default to the original behaviour.

I have now spotted that it'll do a poor job of separating an interleaving of global and datasource options.

gfdatasource/gfdatasource
leth: I think there is currently always only one of them. I'll check this

gfdatasource/gfdatasource
leth: I didn't want to copy the while + sleep loop into both functions, but still avoid duplicate work.

It's probably overcomplicating things.

jml: Maybe just call `cmd_func(grafana_api, opts)` directly, without the temp var?

jml:
LMK when you want a follow-up review.

leth:
@jml I think it's ready for another look :)

PR #63, opened 2017-08-02
core: add Text panel

filippog:


grafanalib/core.py
```diff
@@ -893,6 +894,35 @@ def to_json_data(self):
 
 
 @attr.s
+class Text(object):
```
jml: Please add a docstring, e.g.

```python
"""A Text panel."""
```

grafanalib/core.py
```diff
@@ -893,6 +894,35 @@ def to_json_data(self):
...
+    content = attr.ib()
+    editable = attr.ib(default=True, validator=instance_of(bool))
+    error = attr.ib(default=False, validator=instance_of(bool))
+    height = attr.ib(default=None)
+    id = attr.ib(default=None)
+    links = attr.ib(default=attr.Factory(list))
+    mode = attr.ib(default="markdown")
```
jml: What's the range of values for this? We should probably have a set of constants, as we do for types, evaluators, etc.

filippog: indeed, afaict one of html/markdown/text, I've turned that into constants

filippog:
@jml added changelog entry too, PTAL!

PR #66, opened 2017-08-18
prometheus.py: fix #65 PromGraph can now take tuple or dict as expressions

### LINKED ISSUE #65: Specify target option like intervalFactor in PromGraph
Looking at https://github.com/weaveworks/grafanalib/blob/master/grafanalib/prometheus.py#L8

The `PromGraph` function have a fix call to `Target` to insert `expressions` with only `expr, legend, refId=refId`.

In my case I need to also specify other parameters like `intervalFactor`.

WDYT about change `expressions` `list` of `tuple` arg by `list` of `dict` to allow to specify every `Target` args ?

If needed we could let both behavior compatible by checking if the list contain `tuple` or `dict`.

Something like 

```
    if all(isinstance(expr, dict) for expr in expressions):
        targets = [ 
            G.Target(refId=refId, **args)
            for (args, refId) in zip(expressions, letters)]
    else:
        targets = [ 
            G.Target(expr, legend, refId=refId)
            for ((legend, expr), refId) in zip(expressions, letters)]
```

Could do the job

gaelL:
Related to #65 

Allow to use list of tuple or list of dict as expressions in PromGraph.

This allow to give specific parameters to target if needed, keeping
the possibility to stay on the simple tuple declaration.

### DIFF
--- a/grafanalib/prometheus.py
+++ b/grafanalib/prometheus.py
@@ -12,7 +12,7 @@ def PromGraph(data_source, title, expressions, **kwargs):
         Prometheus data.
     :param title: The title of the graph.
     :param expressions: List of tuples of (legend, expr), where 'expr' is a
-        Prometheus expression.
+        Prometheus expression. Or a list of dict where keys are Target's args.
     :param kwargs: Passed on to Graph.
     """
     letters = string.ascii_uppercase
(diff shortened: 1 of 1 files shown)


jml:
Thanks! I'll take a look at this on Monday.

PR #68, opened 2017-09-06
Add bps and Bps formats

bboreham:
Found [here](https://github.com/grafana/grafana/blob/2fc67da69aab515968d009335389941fbacbb340/public/app/core/utils/kbn.js#L418-L419)


jml:
Thanks!

PR #74, opened 2017-09-15
[WIP]Add more attributes

aknuds1:
As I'm working to convert Prometheus Operator Grafana dashboards to grafanalib, I found that some attributes weren't supported by grafanalib. I add them here.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -13,33 +13,34 @@ from numbers import Number
 import warnings
 
 
-@attr.s
-class RGBA(object):
-    r = attr.ib(validator=instance_of(int))
-    g = attr.ib(validator=instance_of(int))
-    b = attr.ib(validator=instance_of(int))
-    a = attr.ib(validator=instance_of(float))
+def _rgba_ib(default=None):
+    def convert_rgba(val):
+        return 'rgba({}, {}, {}, {})'.format(*val) if val is not None else None
 
-    def to_json_data(self):
-        return "rgba({}, {}, {}, {})".format(self.r, self.g, self.b, self.a)
+    return attr.ib(
+        default=default, validator=instance_of(str), convert=convert_rgba
+    )
 
 
-@attr.s
-class RGB(object):
-    r = attr.ib(validator=instance_of(int))
-    g = attr.ib(validator=instance_of(int))
-    b = attr.ib(validator=instance_of(int))
+def _rgb_ib(default=None):
+    def convert_rgb(val):
+        return 'rgb({}, {}, {})'.format(*val) if val is not None else None
 
-    def to_json_data(self):
-        return "rgb({}, {}, {})".format(self.r, self.g, self.b)
+    return attr.ib(
+        default=default, validator=instance_of(str), convert=convert_rgb
+    )
 
 
-@attr.s
-class Pixels(object):
-    num = attr.ib(validator=instance_of(int))
+def _rgba_list_ib(default=None):
+    def convert_rgba_list(val):
+        return [
+            'rgba({}, {}, {}, {})'.format(*v) for v in val
+        ] if val is not None else None
 
-    def to_json_data(self):
-        return '{}px'.format(self.num)
+    return attr.ib(
+        default=default, validator=instance_of((list, type(None))),
+        convert=convert_rgba_list,
+    )
 
 
 @attr.s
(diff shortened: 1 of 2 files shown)


aknuds1:
@jml How does this look to you, any tests/docs I should update as well?

grafanalib/core.py
```diff
@@ -275,6 +275,10 @@ class XAxis(object):
     def to_json_data(self):
         return {
             'show': self.show,
+            'mode': self.mode,
+            'name': self.name,
+            'values': self.values,
+            'show': self.show,
```
jml: Show is already there. Please remove this second one.

grafanalib/core.py
```diff
@@ -381,8 +385,9 @@ class Row(object):
         default=True, validator=instance_of(bool),
     )
     height = attr.ib(default=DEFAULT_ROW_HEIGHT, validator=instance_of(Pixels))
-    showTitle = attr.ib(default=None)
+    showTitle = attr.ib(default=True)
```
jml: Why are you changing this default?

grafanalib/core.py
```diff
@@ -398,8 +403,9 @@ def to_json_data(self):
             'editable': self.editable,
             'height': self.height,
             'panels': self.panels,
-            'showTitle': showTitle,
+            'showTitle': self.showTitle,
```
jml: I don't understand the motivation for this change. The logic above sets `showTitle` to `True` if someone specifies a title.

grafanalib/core.py
```diff
@@ -714,6 +720,7 @@ class Dashboard(object):
     )
     timezone = attr.ib(default=UTC)
     version = attr.ib(default=0)
+    graphTooltip = attr.ib(default=None, validator=instance_of(int))
```
jml: Why is `graphTooltip` an `int`?

grafanalib/core.py
aknuds1: Probably by mistake, gonna guess I thought I added it myself and that it gave a different result than I thought it should while porting these JSON dashboards.

grafanalib/core.py
aknuds1: Let me dig a bit, there was a reason for it. However, `self.showTitle` isn't used otherwise is it (i.e., it's stale code)?

aknuds1: The reason I made `showTitle` explicit is two-fold:
1. There is an option `showTitle`, which is otherwise ignored
2. I am porting a dashboard which has a hidden title

Is there any reason not to support point 2?

grafanalib/core.py
aknuds1: In the JSON dashboards I'm porting, graphTooltip is an int.

jml: No, I don't. Just asking.

grafanalib/core.py
jml: Fully OK with allowing explicit override of `showTitle`. Just want to keep the current behaviour if it's not specified.

jml:
Sorry, I lost track of the state of this PR. Whose turn is it to look at this?

aknuds1:
@jml Hey interesting you should ask :) Basically, it's been a WIP as I worked on my rewrite of Prometheus Operator to generate Grafana dashboards via grafanalib, but the PR has since been merged!

My thinking now is to clean up my grafanalib fork as it works for the Prometheus Operator dashboards (whereas the output of the original grafanalib does not work), and update this PR. I'll ping you when ready!

jml:
@aknuds1 Hi! Any plans to update this PR?

aknuds1:
@jml Thanks for reminding me, I think I will have the time going forward! Making a note to myself to revisit the PR!

matthewmrichter:
@aknuds1 - We'd be interesting in getting this PR resolved. Is this still WIP? Is it still relevant?

aknuds1:
Hi @matthewmrichter Sorry for never getting back to this PR, I've been very busy since.

How relevant is this library these days? I seem to recall Weaveworks were migrating to Jsonnet instead of Python?

aknuds1:
@matthewmrichter Looking at this PR again, I remember too little of the original context to bring it back into sync with upstream or to remember where I was at in the process. It's best if I close it.

I'll eventually return to it if I need grafanalib again, I've only used jsonnet in the meantime.

PR #77, opened 2017-09-21
Remove hard-coded data source name

bboreham:
This change will require all callers of `QPSGraph` to supply their own name, but I think it's better.


bboreham:
CHANGELOG added

PR #81, opened 2017-10-09
Add Table panel support

mikebryant:


grafanalib/core.py
```diff
@@ -1071,3 +1092,168 @@ def to_json_data(self):
...
+
+    def to_json_data(self):
+        styles = self.styles
+        for column in self.columns:
+            if column.style:
+                style = copy.deepcopy(column.style)
+                if style.pattern and style.pattern != column.text:
+                    raise ValueError("Column style pattern must match the column name if specified")
+                style.pattern = column.text
+                styles.append(style)
```
jml: I don't understand what this is trying to achieve. Could you perhaps provide an example of a `Table` definition that uses styles?

mikebryant:
```
def namespace_usage_panel(decimals, format, query, unit):
    return Table(
        columns=[
            Column(
                style=ColumnStyle(
                    decimals=decimals,
                    unit=format,
                ),
                text=unit,
                value='current',
            ),
        ],
        dataSource='prometheus',
        sort=ColumnSort(col=1, desc=True),
        span=6,
        styles=[
            ColumnStyle(
                alias='Namespace',
                pattern='Metric',
                type='String',
            ),
        ],
        targets=[
            Target(
                expr=query,
                intervalFactor=10,
                legendFormat='{{namespace}}',
                refId='A',
            ),
        ],
        title="%s by Namespace" % unit,
        transform=AGGREGATIONS_TRANSFORM,
    )


...some dashboard definition
        Row(
            collapse=True,
            panels=[
                namespace_usage_panel(
                    decimals=3,
                    format=SHORT_FORMAT,
                    query='namespace:kube_pod_container_resource_requests_cpu_cores:sum_onlyrunning',
                    unit='CPU Cores'
                ),
                namespace_usage_panel(
                    decimals=0,
                    format=BYTES_FORMAT,
                    query='namespace:kube_pod_container_resource_requests_memory_bytes:sum_onlyrunning',
                    unit='Memory'
                ),
            ],
            title='Per-namespace Usage',
        ),

```

mikebryant: Grafana requires specification of the column styles by regex patterns on the column names.

In Python it's much easier in the common case to just specify the column style with the column, instead of having them separated.

grafanalib/core.py
```diff
@@ -1071,3 +1092,168 @@ def to_json_data(self):
...
+                'dateFormat': self.dateFormat,
+            })
+        else:
+            data.update({
+                'colorMode': self.colorMode,
+                'colors': self.colors,
+                'decimals': self.decimals,
+                'thresholds': self.thresholds,
+                'unit': self.unit,
+            })
```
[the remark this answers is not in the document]
mikebryant: Would you prefer I just do 4 different classes for the 4 column styles? It feels like that would be a lot of duplication (Since from the discussion it looks like inheritance would also be a bad idea)

grafanalib/core.py
```diff
@@ -1071,3 +1092,168 @@ def to_json_data(self):
...
+            'col': self.col,
+            'desc': self.desc,
+        }
+
+
+@attr.s
+class Column(object):
+    """Details of an aggregation column in a table panel
+
+    :param style: shorthand for specifying a style that maps to this column
```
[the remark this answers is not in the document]
mikebryant: This should be an instance of a `ColumnStyle` or `None`.

I think I tried to use a `validator` in the `attr.ib` call but it didn't like the default value.. Not sure now.

grafanalib/core.py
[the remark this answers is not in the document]
jml: Not sure. Will think about it & get back to you.

grafanalib/core.py
jml: Thanks for the explanation!

jml:
Thanks for getting back to me @mikebryant. Sorry for the delay in replying.

I'm going to be away until Oct 30, so probably won't have a chance to give this the review it deserves until then.

jml:
I haven't abandoned this. I'm mucking around with it when I get the chance to get my own thinking clear.

jml:
Superseded by #104

PR #83, opened 2017-10-11
Allow specifying a data source input

atopuzov:
* Allows creation of dashboards which parametrise the input datasource [1]

Python code
```
...
dashboard = Dashboard(
    inputs=[
        DataSourceInput(
            name="DS_PROMETHEUS",
            type="datasource",
            label="Prometheus",
            pluginId="prometheus",
            pluginName="Prometheus"
        )],
...
```

Generated JSON
```
...
  "__inputs": [
    {
      "description": "",
      "label": "Prometheus",
      "name": "DS_PROMETHEUS",
      "pluginId": "prometheus",
      "pluginName": "Prometheus",
      "type": "datasource"
    }
  ],
...
```

Python code
```
...
    inputs=[
        ConstantInput(
            name="VAR_PREFIX",
            label="prefix",
            value="collectd"
        ),
...
```

Generated JSON
```
...
  "__inputs": [
    {
      "description": "",
      "label": "prefix",
      "name": "VAR_PREFIX",
      "type": "constant",
      "value": "collectd"
    },
...
```

[1] http://docs.grafana.org/reference/export_import/

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -422,13 +422,9 @@ class Annotations(object):
 
 
 @attr.s
-class Input(object):
+class DataSourceInput(object):
     name = attr.ib()
     label = attr.ib()
-
-
-@attr.s
-class DataSourceInput(Input):
     pluginId = attr.ib()
     pluginName = attr.ib()
     description = attr.ib(default="", validator=instance_of(str))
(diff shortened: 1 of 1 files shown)


jml:
Thanks!

This seems pretty useful, but I'm a bit wary of introducing inheritance into grafanalib. We've managed to avoid it so far and I think that's for the best. 

Do you have ideas about how we could do this without it?

atopuzov:
It is only 2 attributes, so no much duplication. Also attr lib did not play well with inheritance (eg. description had to be copied as it has a default).

I'll just remove inheritance and push the change.

atopuzov:
@jml Pushed the changes without inheritance

PR #88, opened 2017-10-27
core: add decimals attribute to YAxis

fho:
Add a decimals attribute to the YAxis class.
It allows specifying the number of decimal places of the axis
description.


jml:
Thanks!

Issue #75, opened 2017-09-19
Tool for generating grafanalib dashboards from JSON 


jml:
For people getting started with grafanalib, it would be helpful to have a way of converting JSON dashboards into grafanalib ones.

Since grafanalib doesn't yet support everything grafana does, when the tool encounters an unsupported feature, it should fail with a friendly message encouraging the user to file a bug.

atopuzov:
I've tried some brute force [in here](https://github.com/atopuzov/grafanalib/blob/from_json/grafanalib/_parse.py) (needs proper error handling, especially when an unknown feature is encountered)  works for a simple dashboard. Would this be a completely wrong approach?

jml:
Thanks! 

I think this is a great start. Some thoughts:

- The `handleFoo` functions could instead be classmethods (`from_json_data`? `parse_json_data`?) on the respective `Foo` classes
- Now that we've got a JSON generator and a JSON parser, we absolutely should have tests that use [hypothesis](http://hypothesis.readthedocs.io/en/latest/)
- I'm not 100% sure, but I think that rather than having one big if/else to guess the type, we could instead exploit our knowledge of the structure of dashboards, e.g. `Dashboard.from_json_data` knows that the way to parse the value of the `rows` key is to use `Row.from_json_data`—does this make sense?
- To fully close off this ticket, we would also need methods that output the Python definitions of the objects. That ought to be done as a separate change, since it's independent of parsing from JSON.

Let me know if you have any questions. Thanks again for contributing!

atopuzov:
Absolutely should move to classmethods `parse_json_data` (actually started there but moved into one file for convenience)
I was thinking of going the other route too.I think my reasoning was to get called only for `objects` (dicts) and that I don't have to loop trough keys and values. Then I discovered later I need to parse other things too (eg. `RGB`, `px` etc) so no way to avoid going trough keys and values (to detect unknown features, for known it's easy as I now what type of object is under which key). I can try and see how it compares (it should actually be simpler)

atopuzov:
It looks a [bit better](https://github.com/atopuzov/grafanalib/commit/30c7283b2f1fbe81fb422fa6388c208f1e024f14). It's more verbose but there is no *magic*. Still needs work, especially tests.

jml:
Thanks! I just got back from vacation. Will take a look at this soon.

atopuzov:
[...]

Issue #23, opened 2017-03-27
PROPOSAL: classes refactor


dmajere:
This one is intended to start a discussion, here is my proposal:

All the classes must have `to_json_data` method, so it's a good idea to factor it out in a parent class.

It can be something like this:

```
@attr.s
class JsonObject(object):

    def to_json_data(self):
        return dict(
            (a.name, getattr(self, a.name))
            for a in self.__attrs_attrs__ if getattr(self, a.name) is not None)
```

we'll have one major problem with this approach: it seams that there is no hard rule on grafana json fields naming. some fields are in camelcase, some are in snakecase.
So, we can either keep class field names equal to grafana json names, or introduce some kind of mapping, like this:

```
@attr.s
class JsonObject(object):
    FIELD_MAPPING = {
        'dataSource': 'datasource'
    }

    def get_field_mapping(self, field):
        return self.FIELD_MAPPING.get(field, field)

    def to_json_data(self):
        return dict(
            (self.get_field_mapping(a.name), getattr(self, a.name))
            for a in self.__attrs_attrs__ if getattr(self, a.name) is not None)
```

jml:
Interesting! Thanks for proposing this.

Some thoughts:

* I'm going to insist on consistent naming style for Python variables / fields
* I'd like to [avoid subclassing](https://www.youtube.com/watch?v=3MNVP9-hglc) if we can. One way we could do this is to rely on attrs to inspect fields:

    ```python
    def to_json_data(obj):
      return attr.asdict(obj)
    ```

   If we are worried about objects defining attributes that aren't part of the Grafana JSON schema (and we should be), then we can filter based on some metadata (see below). Another way is to have a class decorator, but I think that's unnecessary.
* I'm uncertain about automating this away:
  * PRO: less duplication, thus less scope for silly errors
  * CON: less explicit, which increases scope for difficult-to-understand errors
  * Do you have any more pros & cons?
* attrs allows attributes to [specify a `metadata` field](https://attrs.readthedocs.io/en/stable/api.html#attr.ib), perhaps we can use this to store the Grafana field mapping, e.g.

    ```python
    @attr.s
    class Graph(object):
      dataSource = attr.ib(metadata={'grafanaName': 'datasource'})
    ```
* We could even wrap this in a helper, `grafanaField`, so we can explicitly mark attributes as being exported

What do you think?

dmajere:
I totally agree with keeping fields naming in python consistent.

[...]

PR #91, opened 2017-11-01
[WIP] Generate dashboard code from json

atopuzov:
This is still **WIP**. Use for discussion on direction.

Tries to solve: https://github.com/weaveworks/grafanalib/issues/75
Tested with https://github.com/percona/grafana-dashboards
Had some success at "beautifying" the output with https://pythoniter.appspot.com/


jml:
Thanks! I'm excited about this. I think the approach of parsing JSON into Grafanalib objects and then pretty-printing those is the right one.

I'm going to insist that we have [Hypothesis](http://hypothesis.works/)-powered roundtripping tests for this code. That's going to be a lot of work, I'm afraid. I'm happy to chip in, but I'm unlikely to go as fast as you'd like.

atopuzov:
@jml Can you provide a small example of hypotestis testing and I can take from there (I haven't used it before).

jml:
```python
from hypothesis import given
from hypothesis import strategies as st

def colors():
    """Generate arbitrary value from 0 to 255, for RGB color."""
    return st.integers(min_value=0, max_value=255)

def rgbas():
    """Generate arbitrary, valid RGBA objects."""
    return st.build(RGBA, r=colors(), g=colors(), b=colors(), a=st.floats(min_value=0, max_value=1)

@given(x=rbgas())
def test_rgbas_roundtrips(x):
    """RGBA can be encoded and then decoded."""
    json_dict = x.to_json_data()
    y = RGBA.parse_json_data(json_dict)
    assert x == y
```

I think if you use `classmethod` rather than `staticmethod` you can do `x.parse_json_data`, which would allow you to extract a re-usable roundtripping tester, but my knowledge of Python minutiae is a little stale.

Does that help?

atopuzov:
@jml Awesome! Thanks. Will take it for here as time allows.

atopuzov:
@jml Managed to do some hypotesis testing (which is awesome, just like haskell quickcheck).
Needs a bit of cleanup and making the things generated a bit stricter. How does it look to you?

jml:
Oh wow, this is great, thank you.

Some thoughts:
* I'm kind of fussy about Python formatting. If you could make this match the rest of grafanalib, lovely. If not, I'll happily do the re-indenting myself.
* I'd rather we not use destructive operations like `pop` on things that get passed to us, reserving it for values that only live inside a single scope.
* Rather than using `@staticmethod` and then manually stating the class name (e.g. `return Template(*args, **kwargs)`), why not use `@classmethod` and then `return cls(*args, **kwargs)`?

I'll follow up with some specific in-line comments.

(Also, yes, Hypothesis is awesome. I love Haskell and I love QuickCheck, but Hypothesis is way better, IMHO.)

Makefile
```diff
@@ -66,7 +66,7 @@ lint: $(VIRTUALENV_BIN)/flake8
 	$(VIRTUALENV_BIN)/flake8 gfdatasource/gfdatasource grafanalib
 
 test: $(VIRTUALENV_BIN)/py.test
-	$(VIRTUALENV_BIN)/py.test --junitxml=$(JUNIT_XML)
+	$(VIRTUALENV_BIN)/py.test --junitxml=$(JUNIT_XML) --cov=grafanalib --cov-report term-missing
```
jml: Nice! What do we need to do to get coverage stats as a README shield?

grafanalib/_gen.py
```diff
@@ -124,3 +128,34 @@ def generate_dashboards_script():
...
     """Entry point for generate-dasboard."""
     run_script(generate_dashboard)
+
+
+def parse_dashboard(args):
+    parser = argparse.ArgumentParser(prog='generate-dashboard')
+    # parser.add_argument(
+    #     '--output', '-o', type=os.path.abspath,
+    #     help='Where to write the dashboard JSON'
+    # )
```
jml: Please delete commented out code before merging.

grafanalib/tests/test_parsing.py
```diff
@@ -0,0 +1,444 @@
...
+                     hideZero=st.booleans(),
+                     rightSide=st.booleans(),
+                     sideWidth=unknown(),
+                     sort=unknown(),
+                     sortDesc=unknown())
+
+
+# def targets():
+#     return st.builds(G.Target,
+#                      # target param is usde by st.builds ;-(
```
jml: I asked DRMacIver about this and he suggested these options:

* if `target` is a relatively early argument,  you could pass it positionally (I'm not sure we can do this)
* use `composite` 
* use `builds` with `lambda _target, **kwargs: G.Target(target=_target, **kwargs)`

I have no preference between these.

grafanalib/tests/test_parsing.py
```diff
@@ -0,0 +1,444 @@
...
+    return st.builds(G.Tooltip,
+                     msResolution=st.booleans(),
+                     shared=st.booleans(),
+                     sort=st.integers(),
+                     valueType=st.text(string.printable))
+
+
+def xaxes():
+    return st.builds(G.XAxis,
+                     # XAxis loses information
```
jml: I don't get it. Can you provide more context?

grafanalib/tests/test_parsing.py
```diff
@@ -0,0 +1,444 @@
...
+                     graphTooltip=st.integers())
+
+
+def json_round_trip(obj):
+    """Dumps the object to a JSON and load it back. This is necessary as the
+    DashboardEncoder takes care of nested structures """
+    json_data = io.StringIO()
+    _gen.write_dashboard(obj, json_data)
+    json_data.seek(0)
+    return json.load(json_data)
```
jml: Another approach would be:

```python
def json_round_trip(obj):
    output = json.dumps(dashboard.to_json_data(), cls=DashboardEncoder)
    return json.loads(output)
```

grafanalib/core.py
```diff
@@ -11,6 +11,11 @@
 import math
 from numbers import Number
 import warnings
+import re
+
+
+class ParseJsonException(Exception):
+    pass
```
jml: Any particular reason for raising exception vs returning None? I don't have an opinion on which is better.

grafanalib/core.py
```diff
@@ -266,6 +398,13 @@ def to_json_data(self):
             'value_type': self.valueType,
         }
 
+    @staticmethod
+    def parse_json_data(data):
+        if 'value_type' in data:
+            data['valueType'] = data.pop('value_type')
+
+        return Tooltip(**data)
```
jml: Using this as a concrete example for some of the points mentioned in the top-level comment:

```python
@classmethod
def parse_json_data(cls, data):
    return cls(**rename_keys(data, value_type='valueType'))

def rename_keys(data, **keys):
    new_data = dict(data)
    for old_key, new_key in keys.items():
        if old_key in new_data:
            new_data[new_key] = new_data.pop(old_key)
    return new_data
```

grafanalib/core.py
```diff
@@ -431,6 +598,25 @@ def to_json_data(self):
...
+        return Annotations(**data)
+
+
+def parse_input(data):
+    object_type = data.get('type')
+
+    if object_type == DATASOURCE_TYPE:
+        return DataSourceInput.parse_json_data(data)
+    elif object_type == CONSTANT_TYPE:
+        return ConstantInput.parse_json_data(data)
```
jml: I wonder if we should have a top-level dict:

```python
INPUT_TYPES = {
    DATASOURCE_TYPE: DataSourceInput,
    CONSTANT_TYPE: ConstantInput,
}
```

And then this code would become:

```python
def parse_input(data):
    object_type = data.get('type')
    try:
        return INPUT_TYPES[object_type].parse_json_data(data)
    except KeyError:
        raise ParseJsonException("Unknown input type {}".format(object_type))
```

grafanalib/core.py
```diff
@@ -551,6 +764,27 @@ def to_json_data(self):
...
+        if 'current' in data:
+            current = data.pop('current')
+            if 'text' in current:
+                data['default'] = current['text']
+            elif 'value' in current:
+                data['default'] = current['value']
+        if 'datasource' in data:
+            data['dataSource'] = data.pop('datasource')
+
+        # Deleted values
```
jml: Why are they deleted?

grafanalib/core.py
```diff
@@ -826,8 +1107,49 @@ def to_json_data(self):
...
+    panel_type = panel.get('type')
+
+    if panel_type == GRAPH_TYPE:
+        return Graph.parse_json_data(panel)
+    elif panel_type == TEXT_TYPE:
+        return Text.parse_json_data(panel)
+    elif panel_type == SINGLESTAT_TYPE:
+        return SingleStat.parse_json_data(panel)
+    elif panel_type == TABLE_TYPE:
+        return Table.parse_json_data(panel)
```
jml: See comment on "input types"

grafanalib/core.py
```diff
@@ -826,8 +1107,49 @@ def to_json_data(self):
...
+            data['rows'] = [Row.parse_json_data(row) for row in data['rows']]
+        if 'time' in data:
+            data['time'] = Time.parse_json_data(data['time'])
+        if 'timepicker' in data:
+            data['timePicker'] = TimePicker.parse_json_data(
+                data.pop('timepicker'))
+        if 'links' in data:
+            data['links'] = [DashboardLink.parse_json_data(link)
+                             for link in data['links']]
+
```
jml: This pattern appears in a couple of places. I wonder if we can factor something out.

```python
def parse_json_data(data):
    new_data = transform_dict(
        data,
        __inputs=('inputs', parse_inputs),
        annotations=('annotations', Annotations.parse_json_data),
        ...
    )
    return Dashboard(**new_data)

def transform_dict(d, **keys):
    """Apply functions to individual keys of the dict, optionally renaming them.

    :param dict d: Dictionary to be transformed
    :param **keys: Keyword arguments, one for each key in ``d`` that we want to transform. Values should be tuples like ``(new_name, function)``, where ``new_name`` is the new name and ``function`` is applied to the value. If ``d`` doesn't have a key matching ``old_name`` then ``function`` will not get called.
    """
    new_d = dict(d)
    for old_key, (new_key, function) in keys.items():
        if old_key in new_d:
            new_d[new_key] = function(new_d.pop(old_key))
    return new_d
```

What do you think? I think it's possible that we might want the syntax to be `transform_dict(d, new_key=(old_key, function))` rather than the current `transform_dict(d, old_key=(new_key, function))`. If we did that, should also reverse `rename` above.

grafanalib/core.py
```diff
@@ -1139,5 +1559,95 @@ def to_json_data(self):
...
+            data['mappingTypes'] = [Mapping.parse_json_data(map_type)
+                                    for map_type in data['mappingTypes']]
+        if 'height' in data:
+            data['height'] = Pixels.parse_json_data(data['height'])
+
+        return SingleStat(**data)
+
+
+@attr.s
+class Table(object):
```
jml: There's another PR that adds `Table`, #81. I've been shamefully slow in reviewing it.

grafanalib/_gen.py
atopuzov: This part is actually not done yet ;-)

grafanalib/tests/test_parsing.py
atopuzov: Thanks! Will find a way to solve it.

grafanalib/tests/test_parsing.py
atopuzov: There are 4 params (mode, name, values, show) but only 1 (show) is converted to JSON.
https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L280

grafanalib/tests/test_parsing.py
atopuzov: Yes much simpler ;-)
I just wanted to go trough code that already existed for some reason.

grafanalib/tests/test_parsing.py
atopuzov: Used `composite` seemed more straight forward and very similar to `builds`.

grafanalib/core.py
atopuzov: Yup, refactored.

grafanalib/core.py
atopuzov: There is actually one more option, return the data itself (and this actually makes thing work for a roundtrip).
I went with rising as a signal to the user that something needs to be implemented (that's how I caught `Table`).

Makefile
atopuzov: I just used coverage as a way of figuring out if I got all the classes covered by tests (and thus by parse methods) ;-)
We can probably do some neat stuff. eg. export a html version.

grafanalib/core.py
atopuzov: If you take a look at the code above the values were put in the JSON but no attributes were created (used). 
eg. `sort` is set to `1`.
I fixed that.

grafanalib/core.py
atopuzov: Done.

grafanalib/core.py
atopuzov: I ended up doing it a bit differently:

```
class dicttransform(
        namedtuple('dicttransform', ['old_key', 'new_key', 'transform'])):
    '''Helper data structure used to simplify the transformations on data when
    parsing the object from JSON. Used in conjunction with `transform_dict`
    :param str old_key: Source key
    :param str new_key: Destination key. If not specified same as `old_key``
    :param function transform: Function used to transform the data, if not
                               specified uses identity function
    Example usage:
    dicttransform('old_key', 'new_key') - renames keys
    dicttransform('old_key', transform=some_function)
     - uses some_function to transform data at `old_key`
    dicttransform('old_key', 'new_key', some_function)
     - use some_function to transform the data and rename the key from
       `old_key` to `new_key`
    '''

    def __new__(cls, old_key, new_key=None, transform=lambda x: x):
        if new_key is None:
            new_key = old_key

        return super(dicttransform, cls).__new__(cls, old_key, new_key,
                                                 transform)


def transform_dict(data, *key_transformations):
    '''
    :param dict data: Dictionary to be transformed
    :param list(dicttransform): Transformation to perform
    '''
    new_data = dict(data)
    for old_key, new_key, transform in key_transformations:
        if old_key in new_data:
            old_value = new_data.pop(old_key)
            new_data[new_key] = transform(old_value)
    return new_data
```

As there is a case for:
1. Renaming the key and keeping the value as it was
2. Processing the value (keeping the key name)
3. Both 1 and 2, process the value and rename the key

Implemented in https://github.com/weaveworks/grafanalib/pull/91/commits/a8bd301a393a81bfbdd2e18a8bae15ef7ab709f6
Looks way cleaner.

What do you think?

dogopupper:
Can this be tested / instructions for doing so? @atopuzov

atopuzov:
@dogopupper you can checkout the branch from my repo and pip install it.
eg.
```
git clone git@github.com:atopuzov/grafanalib.git
git checkout from_json
pip install . # I suggest you do it in a virtualenv
parse-dashboard <json-file> -o <python-file>
```
Obviously it is dependent on what is implemented in grafanalib atm.

atopuzov:
@jml I think I have addressed all the concerns you had. Can you give it another look? Thanks!

jml:
Will do.

dogopupper:
@atopuzov @jml having trouble with importlib.machinery SourceFileLoader:

https://docs.python.org/3/library/importlib.html#importlib.machinery.SourceFileLoader


```Traceback (most recent call last):
  File "/usr/local/bin/parse-dashboard", line 9, in <module>
    load_entry_point('grafanalib==0.3.0', 'console_scripts', 'parse-dashboard')()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources/__init__.py", line 565, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources/__init__.py", line 2697, in load_entry_point
    return ep.load()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources/__init__.py", line 2370, in load
    return self.resolve()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/pkg_resources/__init__.py", line 2376, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Library/Python/2.7/site-packages/grafanalib/_gen.py", line 9, in <module>
    from importlib.machinery import SourceFileLoader
ImportError: No module named machinery```

**Comment by dogopupper** [comment]
when using withn python 3.6:

```Traceback (most recent call last):
  File "/Users/dogopupper/work/grafanalib/p35/bin/parse-dashboard", line 11, in <module>
    load_entry_point('grafanalib==0.3.0', 'console_scripts', 'parse-dashboard')()
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 159, in parse_dashboard_script
    run_script(parse_dashboard)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 120, in run_script
    sys.exit(f(sys.argv[1:]))
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 151, in parse_dashboard
    dashboard = grafanalib.core.Dashboard.parse_json_data(json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1213, in parse_json_data
    transform=foreach(DashboardLink.parse_json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in inner
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in <listcomp>
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 642, in parse_json_data
    dicttransform('height', transform=Pixels.parse_json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in inner
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in <listcomp>
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 640, in <lambda>
    transform=foreach(lambda pnl: parse_object(pnl, PANEL_TYPES))
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1756, in parse_object
    return mapping[object_type].parse_json_data(obj)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1663, in parse_json_data
    transform=foreach(ValueMap.parse_json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in inner
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in <listcomp>
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1420, in parse_json_data
    return cls(**data)
TypeError: __init__() got an unexpected keyword argument 'from'```

**Comment by atopuzov** [comment]
@dogopupper Found the bug, will push a fix. Thanks.

**Comment by dogopupper** [comment]
Thank you. New error (3.6):

```Traceback (most recent call last):
  File "/Users/n.penteridis/work/grafanalib/p35/bin/parse-dashboard", line 11, in <module>
    load_entry_point('grafanalib==0.3.0', 'console_scripts', 'parse-dashboard')()
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 159, in parse_dashboard_script
    run_script(parse_dashboard)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 120, in run_script
    sys.exit(f(sys.argv[1:]))
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 151, in parse_dashboard
    dashboard = grafanalib.core.Dashboard.parse_json_data(json_data)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1213, in parse_json_data
    transform=foreach(DashboardLink.parse_json_data)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in inner
    return [func(arg) for arg in args]
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in <listcomp>
    return [func(arg) for arg in args]
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 642, in parse_json_data
    dicttransform('height', transform=Pixels.parse_json_data)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/n.penteridis/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 133, in parse_json_data
    match = Pixels.REGEX.match(data)
TypeError: expected string or bytes-like object```

Would it be best if I passed my 3k lines JSON with multiple graphs to you so you can further develop the solution until it works?

**Comment by dogopupper** [comment]
Thank you @atopuzov . New error (3.6):

```Traceback (most recent call last):
  File "/Users/dogopupper/work/grafanalib/p35/bin/parse-dashboard", line 11, in <module>
    load_entry_point('grafanalib==0.3.0', 'console_scripts', 'parse-dashboard')()
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 159, in parse_dashboard_script
    run_script(parse_dashboard)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 120, in run_script
    sys.exit(f(sys.argv[1:]))
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/_gen.py", line 151, in parse_dashboard
    dashboard = grafanalib.core.Dashboard.parse_json_data(json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 1213, in parse_json_data
    transform=foreach(DashboardLink.parse_json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in inner
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 60, in <listcomp>
    return [func(arg) for arg in args]
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 642, in parse_json_data
    dicttransform('height', transform=Pixels.parse_json_data)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 54, in transform_dict
    new_data[new_key] = transform(old_value)
  File "/Users/dogopupper/work/grafanalib/p35/lib/python3.6/site-packages/grafanalib/core.py", line 133, in parse_json_data
    match = Pixels.REGEX.match(data)
TypeError: expected string or bytes-like object```

Would it be best if I passed my 3k lines JSON with multiple graphs to you so you can further develop the solution until it works?

**Comment by atopuzov** [comment]
@dogopupper  If you can drop it somewhere would be great. Eg. a github gist, pastebin.

**Comment by dogopupper** [comment]
https://pastebin.com/QGbYPL6m 🍰

**Comment by atopuzov** [comment]
@dogopupper I tested with your input and now the only issue is when the row height is `null` or empty string.
Looking at 1 for `Row` it has to be an instance of `Pixels`. There is also `DEFAULT_ROW_HEIGHT` which is 250. `height` is also defined as an attribute on `Graph`, `SingleStat`, `Text` and `Table` but it can be `None` there.I made it so that for `Row` it can be `None` as well through `att.validators.optional` (i should also make it that other object do the same). 
At the moment you need to remove `__requires` from the JSON as `grafanalib` does not have necessary objects to represent that. Unfortunately your dashboard is not accessible anymore and I edited the one I downloaded.

[1] https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L400

**Comment by dogopupper** [comment]
@atopuzov oooooo it works!!!!!!!!!!!! however..... the python result is 20% longer (800 extra lines, around 4100)...... it's like a huge yaml file with the indentation mess (after using the pythoniter beautify)...... is the resulting grafanalib code what is actually needed to produce the dashboard or is there a lot of redundant information that can be further worked on?

**Comment by atopuzov** [comment]
@dogopupper Yes there is a lot of redundant information (eg. defaults). The object is just pretty printed to a file (eg. you would get the same result for any dashboard you define with the minimum of python code and pretty print it to file, it would also be huge).
For the moment I haven't actually thought of a way to deal with that.

**Comment by jml** [comment]
> Yes there is a lot of redundant information (eg. defaults). The object is just pretty printed to a file (eg. you would get the same result for any dashboard you define with the minimum of python code and pretty print it to file, it would also be huge). For the moment I haven't actually thought of a way to deal with that.

FWIW, I'm OK with this still being the case when we merge, and am also OK with doing a release in this state, as long as the command-line tool is marked as experimental.

Once we've done the hard work of turning the JSON into Python objects (thanks again @atopuzov!), manipulating those objects is relatively easy.

File: grafanalib/core.py [hunk 13]
```diff
@@ -596,7 +596,8 @@ class Row(object):
     editable = attr.ib(
         default=True, validator=instance_of(bool),
     )
-    height = attr.ib(default=DEFAULT_ROW_HEIGHT, validator=instance_of(Pixels))
+    height = attr.ib(default=DEFAULT_ROW_HEIGHT,
+                     validator=optional(instance_of(Pixels)))
```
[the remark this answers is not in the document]
atopuzov: Addressed in 512c818e78d867baaddafbcd5b17cae02e2332e3. Also made all `height` attributes either instance of `None` or `Pixels`.

sdementen:
just tested it, great work!
Maybe worth to replace construction of the form as
`       return cls(**new_data)`
by
`       return cls(**{k:v for k,v in new_data.items() if v is not None})`
so that unsupported fields which are null do not break the parsing.

Can this be done on all classes automatically by inheriting some base class ? I am not familiar with the attr package.
More generally, the `to_json_data` is also quite verbose and could benefit from a base_class doing
```
    FIELD_TO_JSON = []
    def to_json_data(self):
        return {k: getattr(self, k) for k in self.FIELD_TO_JSON}
```
and in the children classes having
```
    FIELD_TO_JSON = ["content", "editable", ..., "span"]
```

But my comments could be quite ingenious given my inexperience with grafana, grafanalib and attr...

atopuzov:
@sdementen Hey i remember the authors did not want any inheritance in one of mine other PRs [1]

[1] https://github.com/weaveworks/grafanalib/pull/83

sdementen:
Any comments on why inheritance is frowned upon?

jml:
Hi @sdementen, thanks for commenting!

> Maybe worth to replace construction of the form as
> `return cls(**new_data)`
> by
> `return cls(**{k:v for k,v in new_data.items() if v is not None})`
> so that unsupported fields which are null do not break the parsing.

I don't quite follow. Can you give an example of where this might arise?

> Any comments on why inheritance is frowned upon?

https://www.youtube.com/watch?v=3MNVP9-hglc (summarised in text here: https://www.dencold.com/post/2013/end-of-inheritance-pycon2013/) is probably the best I can do to explain my reluctance.

I'm seriously considering relaxing the restriction on inheritance because we do have some genuine subtypes (e.g. `Graph` and `SingleStat` are both "panels", but this notion isn't represented in the code in any way). Such a change would take place in a different PR—this one is already significant enough.

I probably wouldn't use inheritance in this way to reduce `to_json_data` duplication. Rather than have a base method with a default implementation, I'd prefer a convenience function with that same default implementation, and then have `to_json_data` call that.

jml:
@atopuzov How do we feel about this PR? Is it ready for a final review?

jml:
@sdementen #23 also has some discussion about inheritance.

sdementen:
I used the 
```
        return cls(**{k:v for k,v in new_data.items() if v is not None})
```
in the `Table` class `parse_json_data` method where the new_data dict contained fields not recognized in the `__init__` method but with None values ==> without it, it breaks as unrecognised parameters are sent to `Table.__init__`

jml:
Why were you doing that? How was `new_data` populated? Why did it have fields that weren't attributes of `Table`? 

I don't mean to be defensive: just trying to get some context to better understand how this will help.

sdementen:
I am using the parse-dashboard command to parse a json dashboard extracted from grafana

jml:
OK, so what's going on is that Grafana's JSON has fields for Table that are
not yet supported by grafanalib—is this correct?

sdementen:
yes, correct!

atopuzov:
@jml I need to rebase first and then it's ready for a final review.

atopuzov:
@jml This should be good to go.

jml:
Thanks @atopuzov. I'll take a look as soon as I can. Might not be for a couple of weeks though.

jml:
Hi @atopuzov 

I've got a local rebase of this on my laptop. Is there some way I can share this with you so you don't have to do the laborious work of rebasing yourself?

I had a brief look over the changes and I think I'd like to have a play around with the code you've written, rather than doing a traditional review. 

Here's what I'll be looking into:
- can commit history be cleaned up a bit?
- can we add the Hypothesis stuff (or a simplified version of Hypothesis stuff) _first_ in one PR and then the parser in another?

As well as general exploration to make sure I understand what's going on.

anandsinghkunwar:
Any updates on this?

atopuzov:
Unfortunately I don't have the time ATM. I'll pick it up when time allows. 
Sorry :-(

astub:
What's the status on this?

atopuzov:
Sorry I can't commit any time to this.

jml:
No worries. Thanks for letting us know, and thank you for your efforts!

psmgeelen:
Hi there, I would still love to have this feature, can I be supportive to this?

PR #93, opened 2017-11-02
core: add interval attribute to Target class

fho:
Add the interval attribute to the Target class.
    The attribute is called "Min step" in the grafana GUI.
    
    The default value is an empty string, meaning grafana will determine the
    interval.


jml:
Thank you!

PR #99, opened 2017-12-21
Add support for Elasticsearch datasource

fho:
The PR adds partial support for creating graphs against an elasticsearch datasource.
It only adds support for basic query and aggregators.
Grafana supports more aggregators and query options that are not implemented yet.


fho:
wrong branch

PR #104, opened 2018-01-05
Add table

jml:
Takes #81 and applies all the review feedback that I couldn't articulate without actually writing the code. 

@mikebryant I'd appreciate it if you could take a look. 

The new column style classes are there to help "make invalid states unrepresentable". The rest is there to preserve the idea that we have a bottom layer which maps grafana's data model and then a convenience layer on top.

### DIFF
--- a/gfdatasource/Dockerfile
+++ b/gfdatasource/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.4
+FROM alpine:3.6
 LABEL works.weave.role=system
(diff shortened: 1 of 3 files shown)


mikebryant:
Thanks for looking at it, looks good :)

mikebryant:
@jml Thanks!
(Would we be able to get a release on Pypi with this included soon?)

jml:
I'll try to do a release next week.

PR #107, opened 2018-01-18
Replace `SourceFileLoader` with python2 style `imp.load_source`

teran:
Normally this commit adds python2 support for grafanalib which is
required to work with grafanalib right from `bazel`.

## What does this do?
This commit replaces python3-only way of `importlib.machinery.SourceFileLoader` for dashboards with more generic `imp.load_source` available widely in python2 and python3.

## Why is it a good idea?
This will help to create grafanalib rules for bazel to build dashboards right there without any strange behaviour of various python versions required for particular bazel rules and grafanalib. It will work simply on every python in fact without any impact to announced support for python 3.4 and 3.5.

## Context
Some links about bazel:
 - https://www.bazel.build
 - https://github.com/bazelbuild/rules_python

Some bazel rules like docker ones are written on python and running bazel with specified python version only doesn't work properly since forces all the staff to use the only python which causes break for third-party rules.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,8 +1,8 @@
 machine:
   services:
     - docker
-  python:
-    version: 3.5.2
+  post:
+    - pyenv global 2.7.12 3.4.4 3.5.2
   environment:
     PATH: $HOME/bin:$PATH
     SRCDIR: /home/ubuntu/src/github.com/weaveworks/grafanalib
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   services:
     - docker
   post:
-    - pyenv global 2.7.12 3.4.4 3.5.2
+    - pyenv global 2.7.12 3.4.4 3.5.2 3.6.4
   environment:
     PATH: $HOME/bin:$PATH
     SRCDIR: /home/ubuntu/src/github.com/weaveworks/grafanalib
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   services:
     - docker
   post:
-    - pyenv global 2.7.12 3.4.4 3.5.2 3.6.4
+    - pyenv global 2.7.12 3.4.4 3.5.2 3.6.1
   environment:
     PATH: $HOME/bin:$PATH
     SRCDIR: /home/ubuntu/src/github.com/weaveworks/grafanalib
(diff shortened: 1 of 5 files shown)


jml:
Thanks! I'm happy to extend support to Python 2, and I'm happy to merge this as-is, but, without an update to our CI system, we are inevitably going to make changes that break Python 2 functionality.

Could I tempt you into updating the CI config so we run with tox, and run against 2.7 and a couple of recent 3.Xs? If not, I'll merge this as-is, and Python 2 support will remain best-effort.

jml:
(Also, I really like Bazel. I'd love a PR that showed an example of using grafanalib with Bazel, or even just an example in a comment here.)

teran:
@jml Hello Jonathan, I could try to add some tests but AFAICS you're using CIrcleCI python configuration but tox, so let me try a bit :)

About bazel rules I could post implementation for them or probably even publish rules repository with some free licence, I'll check out which way is possible.

jml:
Thanks!

teran:
@jml So it looks done now and ready to land, about bazel rules I gonna clarify that question with my company today and will post an update here.

Thanks.

teran:
Simple rule to use grafanalib from bazel could look such way:

```
def _grafanalib_impl(ctx):
  args = []
  args.append("-o")
  args.append(ctx.outputs.json.path)
  args.append(ctx.file.src.path)

  ctx.action(
    executable = ctx.executable.generator,
    inputs = [ctx.file.src],
    arguments = args,
    outputs = [ctx.outputs.json],
  )

grafanalib_build = rule(
  implementation = _grafanalib_impl,
  attrs = {
    'src': attr.label(
      allow_files=True,
      single_file=True
    ),
    'generator': attr.label(
        executable = True,
        cfg = 'host',
        default = '@//buildutils/grafanalib:generator',
    ),
  },
  outputs = {
    "json": "%{name}.json",
  },
)
```

with a wrapper:
```
#!/usr/bin/env python

from grafanalib._gen import generate_dashboard_script

if __name__ == '__main__':
    generate_dashboard_script()
```

So it could work the following way:
```
load('//buildutils/grafanalib:grafanalib.bzl', 'grafanalib_build')

grafanalib_build(
  name = "test-board",
  src = "test-board.dashboard.py",
)
```

jml:
Thanks! We might need to do some follow-up work to run tox or similar in the CI, but I think this is good to merge.

PR #108, opened 2018-01-18
Use 'tox' to run tests

jml:
Allows for testing multiple Pythons.

- [x] make sure tests pass
- [x] update changelog
- [ ] use six

Builds on #107

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,56 +23,53 @@ DOCKER_IMAGE_DIRS=$(patsubst %/Dockerfile,%,$(DOCKERFILES))
 IMAGE_NAMES=$(foreach dir,$(DOCKER_IMAGE_DIRS),$(patsubst %,$(IMAGE_PREFIX)/%,$(shell basename $(dir))))
 
 # Python-specific stuff
-VIRTUALENV_DIR ?= .env
-VIRTUALENV_BIN = $(VIRTUALENV_DIR)/bin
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -23,56 +23,53 @@ DOCKER_IMAGE_DIRS=$(patsubst %/Dockerfile,%,$(DOCKERFILES))
...
 
-lint: $(VIRTUALENV_BIN)/flake8
-	$(VIRTUALENV_BIN)/flake8 gfdatasource/gfdatasource grafanalib
+lint: .ensure-flake8
+	$(FLAKE8) gfdatasource/gfdatasource grafanalib
 
-test: $(VIRTUALENV_BIN)/py.test
-	$(VIRTUALENV_BIN)/py.test --junitxml=$(JUNIT_XML)
+test: .ensure-tox
+	$(TOX) --skip-missing-interpreters
```
leth: I can see why you'd want to skip when developing locally, but will they be skipped by CI?

leth:
Regarding six: I suggest you vendor it into the project, rather than adding it as dependency.
The whole world tends to depend on six, which creates a version constraint satisfication nightmare.

jml:
Thanks! We don't yet need six (according to our tests), so I'm skipping that for now.

Makefile
jml: Yes, probably. I'm going to risk it though.

PR #116, opened 2018-02-27
bugfix: SparkLine __init__  error.

xxr3376:
SparkLine use BLUE_RGB as a function, which will lead to `TypeError: 'RGB' object is not callable` error.


jml:
Thanks! I've merged a variant of this (with tests), and released 0.5.1

xxr3376:
@jml Great, thanks for quick response!

PR #119, opened 2018-04-18
Feature/mixed datasources

christine-gong:
## What does this do?

Support the mixed data sources case for Graphs and Alert

## Why is it a good idea?

Currently, there is no datasource attribute set in the Target level, making it impossible to have a Graph with mixed datasources or set the alerts with mixed datasources.
Introducing the datasource attribute in the Target class will help support it

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 =========
 Changelog
 =========
+0.5.3 (unreleased)
+==================
+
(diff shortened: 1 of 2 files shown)


tinder-ubhatnagar:
👍

tinder-ubhatnagar:
@jml - This feature is a must for Prometheus datasources! Please review!

PR #126, opened 2018-05-11
Template: add refresh arg.

gaelL:
Allow to specify the refresh behavior of the template.

  * On Dashboard Load
  * On Time Range Change

Ref : http://docs.grafana.org/reference/templating/#query-options

>Controls when to update the variable option list (values in the dropdown). On Dashboard Load will slow down dashboard load as the variable query needs to be completed before dashboard can be initialized. Set this only to On Time Range Change if your variable options query contains a time range filter or is dependent on dashboard time range.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -177,6 +177,11 @@ SORT_ASC = 1
 SORT_DESC = 2
 SORT_IMPORTANCE = 3
 
+# Template
+REFRESH_NEVER = 0
+REFRESH_ON_DASHBOARD_LOAD = 1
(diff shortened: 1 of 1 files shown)


jml:
Thanks for the PR. I'll try to look at this as soon as I can.

gaelL:
@bboreham 

Values are `0/1/2`

```
<option label="Never" value="number:0">Never</option>
<option label="On Dashboard Load" value="number:1" selected="selected">On Dashboard Load</option>
<option label="On Time Range Change" value="number:2">On Time Range Change</option>```

I will try to take some time today to update the PR and add some constants for this

PR #129, opened 2018-05-25
Add some defaults, add Target.target, Graph.minSpan, Graph.repeat

tharvik:
## What does this do?

The commits should be quite explicits (and small enough).
 * `Target.expr` does nothing with Grafana 5.1.2, Graphite 1.1.3, `Target.target` works with the same syntax.
 * `Graph` can be repeated too, adding support
 * `Template.default` defaults to None for letting Grafana finding the best value
 * `Template.dataSource`, `Graph.dataSource` default to None for using default DataSource

## Why is it a good idea?
Fix a ~bug, support for repeating graphs, some sane defaults.

## Context

 * `repeat`, `minSpan`: http://docs.grafana.org/reference/templating/#repeating-panels

## Questions

Maybe fixing `Target.expr` isn't the right way to do it and we could do a dedicated `GraphiteTarget`, but as is wasn't already done, it went to `core.py`.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -898,6 +898,12 @@ class Dashboard(object):
 
 @attr.s
 class Graph(object):
+    """
+    Generates Graph panel json structure.
+
+    :param minSpan: Minimum width for each panel
+    :param repeat: Template's name to repeat Graph on
+    """
 
     title = attr.ib()
     dataSource = attr.ib()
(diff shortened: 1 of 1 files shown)


jml:
Looks great, thanks!

Do you know if it makes sense for `expr` and `target` to be set on a `Graph` at the same time? I don't think it would change anything for the PR, but it might factor into some other changes I'm considering.

tharvik:
I don't think so, it seems to depend on the DataSource used.

bboreham:
As a Python-ignoramus, is it considered OK to move parameters around (e.g. `query` on `Template`) ?
(It broke some code which assumed it was the second positional parameter)

jml:
> As a Python-ignoramus, is it considered OK to move parameters around (e.g. `dataSource` on `Template`) ?

Generally, no. This was a backwards incompatible change and really should have been flagged as such in the changelog. Mea culpa.

Unfortunately, if you have a thing `f(x, y, z, foo=None, bar=None)` as above, and you want to make `y` optional, you _have_ to move it, because Python doesn't allow `f(x, y=None, z, foo=None, bar=None)`.

You could always make a new thing, `Templatev2`, but that presents a maintenance burden that I'm not ready to deal with yet, and is probably not appropriate for a pre-1.0 project.

PR #143, opened 2018-09-14
Add support for custom variables

genuss:
## What does this do?
Provides proper support for custom template variables

## Why is it a good idea?
Currently custom variables must be manually updated in dashboard settings after import. If you choose a variable in grafana you will see something like that:
image

After this PR it won't be needed. This was tested on grafana 5.1 and 5.2.
This feature respects custom defined options so if they were provided explicitly they will be preferably used. This allows to add some human-readable definition of values (see how it's done in test):
image

## Questions
I'm not an expert in `attrs` and I put my logic in `__attrs_post_init__()`. I'm not sure if this is a proper place for such logic, please comment here.
I preserved backwards-compatibilty and still take `query` parameter as is. However it be constructed from options if they are provided: I'm not sure about this.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -613,6 +613,7 @@ class Template(object):
 
     name = attr.ib()
     query = attr.ib()
+    _current = attr.ib(init=False, default=attr.Factory(dict))
     default = attr.ib(default=None)
     dataSource = attr.ib(default=None)
     label = attr.ib(default=None)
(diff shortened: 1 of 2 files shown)


genuss:
Sorry, made a regression with default values for other types of template variables. Fix in progress.

BenJTucker:
+1
custom variables don't seem to work at all without this fix. Verified on grafana 6.0.2

JamesGibo:
Closes #188

Issue #128, opened 2018-05-25
Automatic refId generator


vthriller:
I'd really like to see one, similarly to now `auto_panel_ids()` currently work: I just don't care about `refId`s most of the time, yet omitting them (just like in that minimal example below) causes all sorts of weird problems on grafana side, like in 5.1.3 where sibling graphs with empty `refId`s render nothing but loading spinner up until query inspector is opened.

```python
dashboard = Dashboard(
        title = 'testing testing 123',
        rows = [
                Row(panels = [
                        Graph(
                                title = 'zomg constants',
                                dataSource = 'prometheus',
                                targets = [
                                        Target(expr = '1'),
                                        Target(expr = '2'),
                                ],
                        ),
                ]),
        ],
).auto_panel_ids()
```

jml:
I think this is a valid request. In the meantime, you might want to look at something like `PromGraph`.

daveworth:
I've given this enhancement a shot and it wasn't too bad - I hope folks like it and it helps them.

JamesGibo:
Added in #144

PR #144, opened 2018-10-24
add `Graph#auto_ref_ids`

daveworth:
## What does this do?
Adds a `#auto_ref_ids` method to Graph to add `refId` values for those `Target`s which do not have explicit `refId`s set, and preserves the IDs which are specified.  Its implementation is based on `Dashboard`'s `#auto_panel_ids` method.

## Why is it a good idea?
It is requested in #128 and reduces toil for Dashboard creators and maintainers.

## Context
Any target list must have explicit `refId` attributes set today to make Grafana happy.  This is often an "auto-incrementing" alphabetical ID manually maintained by the creator of a dashboard.  This is only useful when one `Target` actually references another one - in other cases it is only toil.  I would like to reduce that toil.

## Questions
None for now!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ TBA
 * Add ``PieChart`` class for creating Pie Chart panels
 * Add `transparent` setting to classes that were missing it (Heatmap, PieChart)
 * Add InfluxDB data source
+* Add ``auto_ref_ids`` to ``Graph``s
 * ...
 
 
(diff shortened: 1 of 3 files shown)


bboreham:
Sorry this languished so long.

It looks like there are two unrelated changes in this PR - one bugfix and one feature - and I fixed the bug in a different way in 3b7b5e16855acb8b46f9a2635b1636844f461355.

The changelog entry needs to be fixed up.  I think I'll do a release and then come back to this.

JamesGibo (approved):
Looks good

JamesGibo:
Re-based on master to fix merge conflicts

dholbach (approved):
Super!

PR #146, opened 2018-10-26
Removing target restrictions.

### LINKED ISSUE #145: Cloudwatch datasources not supported due to Target class restrictions
There is no support for Cloudwatch since the Target class has restrictions. If you remove them, you'll be able to make Cloudwatch graphs.

buroa:
Fix #145

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -281,40 +281,13 @@ class Legend(object):
         }
 
 
-@attr.s
 class Target(object):
-    """
-    Metric to show.
 
-    :param target: Graphite way to select data
-    """
-
-    expr = attr.ib(default="")
-    format = attr.ib(default=TIME_SERIES_TARGET_FORMAT)
-    legendFormat = attr.ib(default="")
-    interval = attr.ib(default="", validator=instance_of(str))
-    intervalFactor = attr.ib(default=2)
-    metric = attr.ib(default="")
-    refId = attr.ib(default="")
-    step = attr.ib(default=DEFAULT_STEP)
-    target = attr.ib(default="")
-    instant = attr.ib(validator=instance_of(bool), default=False)
-    datasource = attr.ib(default="")
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
 
     def to_json_data(self):
-        return {
-            'expr': self.expr,
-            'target': self.target,
-            'format': self.format,
-            'interval': self.interval,
-            'intervalFactor': self.intervalFactor,
-            'legendFormat': self.legendFormat,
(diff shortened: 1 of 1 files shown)


buroa:
Looks like this pipeline is broken, and the tests themselves passed.

```
touch .ensure-pip
/home/ubuntu/virtualenvs/venv-system/bin/flake8 gfdatasource/gfdatasource grafanalib
grafanalib/validators.py:36:12: W605 invalid escape sequence '\d'
grafanalib/validators.py:38:27: W605 invalid escape sequence '\d'
make: *** [lint] Error 1

cd $SRCDIR; make all returned exit code 2```

**Comment by bboreham** [comment]
Hi, and thanks for the PR.  Sorry it took so long to review.

I'm not sure I understand your aim - it seems you have in mind a type of Target that does not match the restrictions coded by previous users.  Rather than deleting every check, it would be more in keeping with the style of this library to introduce a different class tailored to what you need.

Given #145 this might be a `CloudwatchTarget` ?

Given a more detailed rationale we might accept an `ArbitraryTarget`.

**Comment by diogomatsubara** [comment]
Given that this could also fix #122 maybe ArbitraryTarget is a good idea.

**Comment by bboreham** [comment]
Why would you prefer that over an `InfluxDBTarget` that checked relevant parameters?

**Comment by diogomatsubara** [comment]
ArbitraryTarget would be ready for use on this PR while InfluxDBTarget would still have to be implemented. That's the only reason. If InfluxDBTarget is in the plans as a fix for #122 that would be even better.

**Comment by bboreham** [comment]
None of them are implemented. 

You are free to have such a subclass in your own code, but I don’t think it belongs in this library as it doesn’t match the overall style.

**Comment by johanfleury** [comment]
I don't think this PR nor the recommendation of @bboreham to create a `AbitraryTarget` is usefull given that you can already pass arbitrary data to the `targets` field:

```
$ cat test.py
import grafanalib.core as grafana


dashboard = grafana.Dashboard(
    title="Foobar",
    rows=[
        grafana.Row(
            panels=[
                grafana.Graph(
                    title="Foobar",
                    dataSource="influxdb",
                    targets=[
                        {
                            'foo': 'bar',
                        },
                    ],
                )
            ]
        )
    ],
)
$ generate-dashboard test.py
{
[SNIPPED]
          "targets": [
            {
              "foo": "bar"
            }
          ],
[SNIPPED]
}
```

It would be better to have a specific `Target` for each data source.

bboreham:
> the recommendation of @bboreham to create a AbitraryTarget

I said "we might accept" it - that isn't a recommendation.

Anyway, this hasn't moved forward in a while so closing.

PR #149, opened 2019-01-04
Added microsecond, and ensure ascii if special chars - WIP

shubhamagarwal1993:
## What does this do?
 - You can use microsecond as a unit on yaxes
 - The symbol `mu` does not stay as an ascii character, and using `--ensure_ascii` keeps it that way

## Why is it a good idea?
 - the json file for grafana expects `mu` to be the special symbol, but as of now it gets converted to unicode. To not change the existing functionality, only adding `--ensure_ascii` will take care of conversion.

## Context
 - Using the example from the readme, change `frontend.dashboard.py` to take MICROSECONDS instead of SECONDS as follows `yAxes=single_y_axis(format=MICROSECONDS_FORMAT),`
 - Run `generate-dashboard -o frontend.json frontend.dashboard.py` and check the output in `frontend.json`. You will see unicode `\u03bcs` in the y-axes.
 - Now run `generate-dashboard -o frontend.json frontend.dashboard.py --ensure_ascii` and you should see "μs" in y-axes. Uploading this file on grafana should set y-axes to microseconds.

## Questions
 - Do i need to update the CHANGELOG? I tried looking at previous pull requests made and some of them don't have it changed.
 - The README would have to be changed to show `--ensure_ascii` can be used.


shubhamagarwal1993:
@weaveworks-admin-bot master is failing tests, any help would be appreciate. thanks

bboreham:
Hi, thanks for the PR.

I find the logic a bit hard to follow: ASCII is a 7-bit code, and your objective is to embed a character `μ` which is well outside of that range.  Also you are setting the parameter to false when applied; it turns off the default behaviour of ensuring that output is ASCII.  So the option should not be named `ensure-ascii`

I don't really see a problem in hard-coding `ensure_ascii` to `False` in `write_dashboard()` with no option - would that work for you?

bboreham:
No update in a while - closing.  Please reopen if you come back to it.

PR #160, opened 2019-04-15
Update versions of Python tested

bboreham:
Currently CI is only running Python 2.7; it seems like a good idea to run it against the newest version in common with the CircleCI 1.0 config and also the newest version as of today.

I used `pyenv` to run the installs, since this is what the CircleCI 1.0 config did.

EDIT: the upstream build image is now installing Python 3, which requires a couple more changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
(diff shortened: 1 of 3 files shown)


bboreham:
OK, I don't know how to install Python.  If you're reading this and you do know, please post a comment.

PR #161, opened 2019-05-05
Add "for" param for alerts, Grafana 6.X

2easy:
This parametr sets a grace period during which no notifications
are set.

## What does this do?
In Grafana 6.X (probably 5.X too), ``for`` parameter for alert sets grace period during which Grafana does not send notifications about firing alert.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+0.5.3 (2019-05-05)
+==================
+
+Changes
+-------
+
+* Add ``for`` parameter for alerts on Grafana 6.X
+
 0.5.2 (2018-07-19)
 ==================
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+0.5.3 (2019-05-05)
```
bboreham: This change would be after 0.5.3 which was released last year.
(Sorry 0.5.3 didn't make it into the changelog)

Atem18:
Hi @bboreham and @2easy , can you fix it so it can be merged or should I propose another pull request with the rebase ?

bboreham:
Looks like release 0.5.3 was never merged to master so I've done that post-hoc.

Atem18:
@bboreham Nice, thanks for the merge.

PR #162, opened 2019-05-30
Update core.py

srdey3993:
## What does this do?
Add support for GridPosition to specify size or position of a specific panel.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1,5 +1,4 @@
 """Low-level functions for building Grafana dashboards.
-
 The functions in this module don't enforce Weaveworks policy, and only mildly
 encourage it by way of some defaults. Rather, they are ways of building
 arbitrary Grafana JSON.
@@ -181,9 +180,6 @@ SORT_IMPORTANCE = 3
 REFRESH_NEVER = 0
 REFRESH_ON_DASHBOARD_LOAD = 1
 REFRESH_ON_TIME_RANGE_CHANGE = 2
-SHOW = 0
-HIDE_LABEL = 1
-HIDE_VARIABLE = 2
 
 
 @attr.s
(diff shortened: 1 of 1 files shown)


bboreham:
Please tell us a bit more about what you are trying to achieve.

srdey3993:
Hi,

I am trying to add support for GridPosition to specify size or position of
a specific panel.

Thanks,
Sudhanshu

grafanalib/core.py
```diff
@@ -608,7 +618,7 @@ class Template(object):
         :param type: The template type, can be one of: query (default),
             interval, datasource, custom, constant, adhoc.
         :param hide: Hide this variable in the dashboard, can be one of:
-            SHOW (default), HIDE_LABEL, HIDE_VARIABLE
+            0 (default, no hide), 1 (hide label), 2 (hide variable)
```
bboreham: Why this change?

grafanalib/core.py
```diff
@@ -853,7 +860,7 @@ def to_json_data(self):
 class Dashboard(object):
 
     title = attr.ib()
-    rows = attr.ib()
+    struct = attr.ib()
```
bboreham: This will break every existing user of Dashboard, right?
Is there a way to add what you want without breaking existing things?

grafanalib/core.py
```diff
@@ -1,5 +1,4 @@
 """Low-level functions for building Grafana dashboards.
-
```
bboreham: Try to minimize irrelevant changes, like removing a blank line.  They make it harder for people to see what really changed.

bboreham:
Hi, are you likely to come back to this?

srdey3993:
No Bryan, Thanks.

PR #164, opened 2019-06-10
Add 'diff', 'percent_diff' and 'count_non_null' as RTYPE

vicmarbev:
## What does this do?

Added 'diff', 'percent_diff' and 'count_non_null' as RTYPE for AlertConditions

## Why is it a good idea?

It allows grafanalib to use those reducers in AlertConditions

## Context

These three RTYPEs were missing in core.py

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,12 +3,13 @@ Changelog
 =========
 
 Next release
-==================
+=======
 
 Changes
 -------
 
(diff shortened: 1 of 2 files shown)


vicmarbev:
Done!

bboreham:
I think something went wrong with your rebase - there are now 13 commits in this PR.

vicmarbev:
I think now it is correct

bboreham:
Thanks!

PR #165, opened 2019-06-12
Remove trailing spaces from separators

cPu1:
## What does this do?
This removes trailing spaces from separators (`:` and `,`) in the generated JSON.

## Why is it a good idea?
A trailing space, specially after the item separator (`,`), causes Go's YAML marshaling to escape any JSON data embedded inside YAML, instead of preserving line breaks and other escape sequences (using `|`).

This doesn't affect Python 3.4 and above where the default behaviour is changed to not include a whitespace after `,`:
>Changed in version 3.4: Use (',', ': ') as default if indent is not None

### DIFF
--- a/grafanalib/_gen.py
+++ b/grafanalib/_gen.py
@@ -43,7 +43,7 @@ class DashboardEncoder(json.JSONEncoder):
 def write_dashboard(dashboard, stream):
     json.dump(
         dashboard.to_json_data(), stream, sort_keys=True, indent=2,
-        cls=DashboardEncoder)
+        separators=(',', ':'), cls=DashboardEncoder)
     stream.write('\n')
 
 


bboreham:
I don't know enough about Python to get a feel for whether this is a good idea.
Can we just recommend that people use a Python newer than March 16, 2014?

lesinigo:
FYI Python-3.4 seems to be out of support: https://devguide.python.org/#status-of-python-branches

The exact date has not yet been specified in [PEP-429](https://www.python.org/dev/peps/pep-0429/) but I guess it will be sometime later this year. Next year Python-3.5 is also going out of support.

bboreham:
@lesinigo thanks, but we're saying "3.4 and above", and "newer than".

bboreham:
I'm going to close this - people who have this problem should use a newer Python.

PR #168, opened 2019-07-18
close #154 add ok for alert state

### LINKED ISSUE #154: add OK state as an option in alert when alert fails or no data found
Grafana has 4 options while grafanalib has only 3 (here https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L112)

butlerx:
## What does this do?

Adds support for OK state in alerts

## Why is it a good idea?

allows users to treat alerts that havbe no data as ok

## Context

issue #154


bboreham:
Thanks for the PR!

PR #180, opened 2019-10-01
pin to attrs 19.2 and fix deprecated arguments

kevingessner:
## What does this do?
fixes method arguments that were deprecated in attrs, and removed intoday's attrs 19.2 release

## Why is it a good idea?
builds are broken at master right now


cherrot:
@pmaterer Could you please accept this PR and release a new version?  For now I have to run `pip3 install attrs==19.1.0` before installing `grafanalib` to bypass this issue.

Thanks in advance.

PR #183, opened 2019-10-17
Use `converter` on attr.ib()

johnathanbell-slalom:
`convert` was deprecated in #307 of attrs
https://github.com/python-attrs/attrs/issues/307

## What does this do?

Fixes broken attr.ib() calls referencing a deprecated kwarg

## Why is it a good idea?

Installations of grafanalib do not work with newest attrs lib

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,8 @@ Next release
 Changes
 -------
 
-TBA
+* use ``converter`` instead of ``convert`` on ``attr.ib()``
+    * https://github.com/python-attrs/attrs/issues/307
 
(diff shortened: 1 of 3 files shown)


darrylbalderas:
Just checking in to see what is the progress of merging this PR @jml @leth

gruzewski:
You could also update `setup.py` with:
```
    install_requires=[
        'attrs>=19.2.0',
    ],
```

bboreham:
This appears to be a duplicate of #180

johnathanbell-slalom:
Works for me. Didn’t see #180 when I put this in. Thanks!!

PR #185, opened 2019-10-20
Repair tests

uritau:
## What does this do?
This PR makes some small changes in order to pass the test suite.

## Why is it a good idea?
It doesn't change "anything" only some deprecation/errors during tests.

 I think that it's clear to have different changes in splited commits to review the PR so Before merge it will be necessary to squash all the commits.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,10 +57,10 @@ percentile latency:
               refId='E',
             ),
           ],
-          yAxes=[
+          yAxes=G.YAxes(
             YAxis(format=OPS_FORMAT),
             YAxis(format=SHORT_FORMAT),
-          ],
+          ),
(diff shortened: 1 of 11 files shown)


bboreham:
The last CI build passed all tests, so please can you explain why they need fixing.

(Assume I know nothing about Python - see #160)

bboreham:
OK I see your first commit fixed the same thing as #180.  
The rest are fixing warnings, maybe?

bboreham:
Thanks!

dholbach:
@uritau Can you take a look at https://github.com/weaveworks/grafanalib/pull/197 maybe? Looks like the change in `README.rst` broke the example?

PR #172, opened 2019-08-13
rows as panels in dashboard; gridpos for row, graph, text panels

victoriaalee:
## What does this do?
Rows are now considered as panels in the dashboard. The attributes of the `Row`, `Graph`, and `Text` panels are also updated; the main change is adding `GridPos`.

## Why is it a good idea?
Rows as panels and `GridPos` were introduced several dashboard schema versions ago.

## Context
I created a `Panel` class, which the `Row`, `Graph`, and `Text` panels subclass.

Grafana source code referenced:

- https://github.com/grafana/grafana/blob/v6.0.0/public/app/features/dashboard/state/DashboardModel.ts
- https://github.com/grafana/grafana/blob/v6.0.0/public/app/features/dashboard/state/PanelModel.ts
- https://github.com/grafana/grafana/blob/v6.0.0/public/app/plugins/panel/graph/module.ts
- https://github.com/grafana/grafana/blob/v6.0.0/public/app/plugins/panel/text/module.ts

## Questions
N/A

PR #193, opened 2019-11-18
Remove gfdatasource from the repo

### LINKED ISSUE #76: Push gfdatasource to docker hub as well as quay.io
The [blog post](https://www.weave.works/blog/grafana-dashboards-as-code/) refers to weaveworks/gfdatasource:latest, but we aren't pushing there any more. We should keep pushing there because:

- it keeps the blog post relevant
- people look for things on DockerHub

CC @errordeveloper

### LINKED ISSUE #94: Update docs re data source for new grafana releases
ISTR new grafana releases have better ways of specifying the data source. We should find those and refer to them in the docs, promoting gfdatasource only for older versions.

### LINKED ISSUE #140: Don't use build-tools
grafanalib doesn't benefit much from using Weaveworks' common build-tools, as it doesn't need to generate images (now that gfdatasource is obsolete), and it is not built with Go.  We should just stop using it.

bboreham:
As noted in #94, `gfdatasource` is unnecessary as the feature is built in to Grafana.

Since we don't need any tools except Python now, use the Circle CI Python image instead of a Go image.

Fixes #94 
Fixes #140 
Fixes #76


dholbach:
This looks nice and clean and things continue to work. A couple of questions:

- a search on Github shows that a couple of folks use the image - do we need to do anything?
- as #140 gets closed, can we drop the /tools directory?
- will we go with 0.6.0 or will we aim for a 1.0 release (note there's a milestone with a few issues - not sure if we want to fix those all)

@matthewmrichter Your input would be welcome here too.

bboreham:
I think the mentions elsewhere on GitHub stem from using the image in a test for the FluxCD project.  That test was removed some months ago.

The script `tools/image-tag` is called from the Makefile.  Otherwise I think we can drop it.

I think 0.6 is better, since there are other breaking changes up for debate.

matthewmrichter:
I do agree that there are some breaking changes on the horizon that might merit a major release number bump. #172, etc (or equivalent functionality)

dholbach:
Re: #140 - are we going to drop `/tools` in this PR now?

PR #195, opened 2020-01-04
Added more YAxis formats, added Threshold and SeriesOverride types

larsderidder:
## What does this do?

As in the title, just adding of some data formats.

## Why is it a good idea?

It makes the library more complete.

## Context

I needed it for my own implementation where I first define all types as objects, and then serialize them all. In the examples I saw that for at least the seriesOverride, a plain dictionary was used as well, but that didn't suit my use case (plus that it is not as self evident which fields are there).

## Questions

I did not check the added types for completeness, so I might miss fields. Not sure if that's an issue at this stage.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -103,6 +103,8 @@ Changes
 * Add ``STATE_OK`` for alerts
 * Add named values for the Template.hide parameter
 * Add cardinality metric aggregator for ElasticSearch
+* Add Threshold and Series Override types
+* Add more YAxis formats
 
 Many thanks to contributors @kevingessner, @2easy, @vicmarbev, @butlerx.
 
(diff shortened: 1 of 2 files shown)


dholbach:
Just a quick PSA from me: 
1. we're hanging out in [#grafanalib on Slack](https://weave-community.slack.com/archives/grafanalib)
1. we're planning a contributors meeting soon: here's the [link to the doodle](https://doodle.com/poll/hgsp9wmuff7m9bi7) to find a good time - be sure to pick your timezone before submitting ([preliminary agenda](https://docs.google.com/document/d/1JxrSszyPHYhNbJDWYZehRKv6AO4U-zIBhuNmYQVOIHo/edit))

We'd love to see you there.

dholbach:
It looks like the PR was approved, but `master` has moved on since then. Do you think you could rebase this on current?

matthewmrichter:
Can you please rebase this off the latest master to overcome the merge conflict?

PR #197, opened 2020-01-14
Fix example in README

alessandrotrisolini:
## What does this do?
Fix the example in README.rst to make it correctly generate dashboard JSON file.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ Next release
 Changes
 -------
 
-TBA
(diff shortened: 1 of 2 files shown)


dholbach:
Thanks for the fix @alessandrotrisolini - your suggested change makes the example run of `generate-dashboard` pass. I just pinged @uritau to have a quick look.

alessandrotrisolini:
Thanks for the feedback @dholbach - appreciate it!

PR #199, opened 2020-02-14
Fix AlertList panel generation

number492:
## What does this do?
The generation of the AlertList panel was broken, this commit fixes that.

## Why is it a good idea?
Otherwise AlertList panels can't be generated.

## Context
A property was missing, this commit adds it. That's all the context I have.

## Questions
Why did it broke? Did it ever work?


matthewmrichter:
This seems pretty low impact/risk - we will check it shortly

number492:
Thanks for the approve! What should I do to get this merged? I don't have write access.

dholbach:
Can you fix the conflict, and we'll merge it for you. Sorry.

PR #201, opened 2020-02-14
Create release workflow definition

dholbach:
Following https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ in order to make releases happen automatically (#56)

The first test ran against test.pypi.org: https://github.com/dholbach/grafanalib/runs/446202668

Please review and let me know if this generally makes sense.

PR #202, opened 2020-02-14
release 0.5.5

dholbach:
I updated setup.py and the release changelog. Together with #201 we can see if this process works.

Might be worth releasing this on Monday instead? :-) (I'll be off on a :steam_locomotive: in a bit.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,14 +2,26 @@
 Changelog
 =========
 
-Next release
+0.5.5 (2020-02-17)
 =======
+
+It's been a while since the last release and we are happy to get this one into your hands.
+0.5.5 is a maintenance release, most importantly it adds support for Python > 3.5.
+
+We are very delighted to welcome Matt Richter on board as maintainer.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -2,13 +2,23 @@
...
 
 * Update README.rst to make the example work
+* Bump Dockerfile to use Alpine 3.10 as base
+* Fix up ``load_source()`` call which doesn't exist in Python 3.5
+* Update versions of Python tested
+* Repair tests
+* pin to attrs 19.2 and fix deprecated arguments
+
+
+We are happy to add Matt Richter as maintainer - thanks a lot for stepping up!
```
errordeveloper: The construct "happy to do something" is a typically used as a confirmation of one's ability and willingness to do something; something like "We are very delighted to welcome Matt Richter on board as maintainer" would be more appropriate in this context.

dholbach:
Thanks, this makes sense! Committed and rebased on master.

PR #209, opened 2020-03-04
A lower-case `t` was preventing Table time column from formatting properly

matthewmrichter:
## What does this do?
This changes the Table class's default ColumnStyle for "Time" to use upper case `T` instead of lower case `t` in its matching pattern. 

## Why is it a good idea?
In my scenario at least, the Timeseries data I got back had its time column labeled as `time` not `Time` so the time formatting was not being applied. 

## Questions
Is it universal that the data source will say capital T `Time` or does that vary? Should we cover both scenarios?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Changes
 * Removed gfdatasource - feature is built in to Grafana since v5.
 * Generate API docs for readthedocs.org
 * Fix AlertList panel generation
+* Fix lowercase `"time"` pattern for default time_series column format in Table class
 
 0.5.5 (2020-02-17)
 ==================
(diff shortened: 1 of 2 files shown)


bboreham:
If I'm following the history properly the code came from a 3rd-party contributor and was originally lowercase: https://github.com/weaveworks/grafanalib/pull/81/files#diff-f72bea905993b5b58182a641642e0602R1218

So I guess it comes back differently in different situations?
I've never used this feature personally.

Seems the parameter is configurable, so couldn't you make use of that where you use it?

matthewmrichter:
I added another complete `ColumnStyle` object for the lowercase `t` - seems to me this would safely cover both situations.

Issue #55, opened 2017-07-27
Automatically test examples


jml:
It's a bit of a pain to manually test the thing in the docs

dholbach:
Once we fix #4 we could go with https://quick-sphinx-tutorial.readthedocs.io/en/latest/code.html#include-source-code

PR #220, opened 2020-03-23
Test examples automatically

dholbach:
- Add python3.8 (current on Ubuntu 20.04) to tox and CircleCI
- decouple main github README from grafanalib docs RTD where we can use sphinx' "literalinclude"
- fix flake linter issues
- make examples run at all
- test examples as part of general tests

fixes: #55

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,122 +11,43 @@ so, grafanalib is for you.
 
 grafanalib lets you generate Grafana dashboards from simple Python scripts.
 
-Writing dashboards
-==================
+How it works
+============
 
-The following will configure a dashboard with a single row, with one QPS graph
-broken down by status code and another latency graph showing median and 99th
-percentile latency:
+Take a look at `the examples directory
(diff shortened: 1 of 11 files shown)


README.rst
```diff
@@ -11,122 +11,43 @@ so, grafanalib is for you.
...
 grafanalib lets you generate Grafana dashboards from simple Python scripts.
 
-Writing dashboards
-==================
-
-The following will configure a dashboard with a single row, with one QPS graph
-broken down by status code and another latency graph showing median and 99th
-percentile latency:
-
-.. code-block:: python
```
matthewmrichter: I'm a little concerned about removing any example from the top level README. I think it would be good to keep at least a simple example for the benefit of people evaluating options.

dholbach: I agree that it'd be nice, unfortunately the `literalinclude` bit won't work on Github - it'll work on readthedocs though: https://github.com/weaveworks/grafanalib/pull/220/files#diff-22ab0e1b24df9ee7410dc1959a86d339R21

PR #232, opened 2020-04-07
Add Elasticsearch derivative metric aggregator

DWalker487:
## What does this do?
This implements the elasticsearch derivative metric aggregator. 
As it is a pipeline aggregation as it relies on the prescence of other metric aggregations to function.
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-derivative-aggregation.html

## Why is it a good idea?
This allows users to plot the time derivative (change over time) of a given metricaggregator. 

For example, this could be used to plot the change over time of the average value of a metric.

## Context
To use this pipeline aggregator, one must also know the ID of the metric aggregator that it applies to, so this also requires ids to be in place for all of the other elasticsearch metric aggregators. It is also useful to be able to hide the other metric aggregations (see PR #231). The pipleline ID is defined in case multiple such pipeline aggregations are desired.

The need from this has come from wanting to plot the change in metrics that are stored only cumulatively by an Elasticsearch datasource.


CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Changes
 * Add average metric aggregation for elastic search
 * Bugfix to query ordering in Elasticsearch TermsGroupBy
 * Added all parameters for StringColumnStyle
+* Add derivative metric aggregation for Elasticsearch
```
matthewmrichter, line 28: Can I have you rebase this and move this comment up to the "next version" section?

matthewmrichter:
@dholbach maybe you can help with these docs errors?

dholbach:
It's #251. Not sure where those docs moved. Maybe just remove the links? 🤷‍♂️

dholbach:
I pinged Marcus about them on slack too.

dholbach:
Apart from the links, this test fails too:
```
/opt/hostedtoolcache/Python/3.8.3/x64/bin/flake8 grafanalib
grafanalib/elasticsearch.py:139:13: E123 closing bracket does not match indentation of opening bracket's line
```

dholbach:
I created https://github.com/weaveworks/grafanalib/pull/252 as a temporary measure. :eyes: welcome.

PR #233, opened 2020-04-13
Add Data links support

ljades:
## What does this do?
Add data links support to ``Graph``, ``BarGuage``, and ``GuagePanel`` panels

## Why is it a good idea?
Data links are a fantastic feature new to Grafana 6.3, allowing for easy linking of panel datapoints to usable URLs.


matthewmrichter:
Has this been tested with pre-6.3 by chance? I believe it would ignore the option attribute, but I'm not 100% on that.

ljades:
> Has this been tested with pre-6.3 by chance? I believe it would ignore the option attribute, but I'm not 100% on that.

It has not been. What is the best way to efficiently test on earlier versions of Grafana to the standards of this package?

matthewmrichter:
We're running 0.6.3, we will test it.

PR #235, opened 2020-04-16
Sum metric agg

DWalker487:
## What does this do?
This adds support for the Elasticsearch Sum metric aggregator
https://www.elastic.co/guide/en/elasticsearch/reference/7.6/search-aggregations-metrics-sum-aggregation.html

## Why is it a good idea?
This is one of the core elasticsearch metric aggregations, and is natively supported by Grafana. 

## Context
The need for this arose when wanting to combine multiple documents reporting at the same interval to generate a total. This metric can also be hidden, as per #231.


dholbach:
Can you please resolve the conflict in CHANGELOG.rst?

dholbach:
This seems to lead to #245 - do you think you can check?

```console
[daniel@reef grafanalib ]$ python -c 'from grafanalib.elasticsearch import AverageMetricAgg; AverageMetricAgg().to_json_data()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/daniel/dev/grafanalib/grafanalib/elasticsearch.py", line 78, in to_json_data
    'id': str(self.id),
AttributeError: 'AverageMetricAgg' object has no attribute 'id'
[daniel@reef grafanalib ]$ 
```

DWalker487:
Ah yes - I missed an instance where I set up a default id when splitting out commits into different PRs - I'll update this now.

Issue #224, opened 2020-03-23
grafanalib/elasticsearch.py:197: DeprecationWarning: assoc is deprecated and will be removed after 2018/01.


dholbach:
```
  /home/daniel/dev/grafanalib/grafanalib/elasticsearch.py:197: DeprecationWarning: assoc is deprecated and will be removed after 2018/01.
    return attr.assoc(self, bucketAggs=list(map(f, self.bucketAggs)))
```

PR #237, opened 2020-04-29
Use attr.evolve instead of attr.assoc

dholbach:
Apparently `attr.assoc` has been replaced by `attr.evolve` since a long time ago: https://www.attrs.org/en/stable/api.html#attr.assoc

closes: #224

Applying the PR and running `tox` won't show the warning mentioned in #224


dholbach:
Log messages look :+1:  to me.

PR #238, opened 2020-04-29
move to GH actions

dholbach:
I got the feeling that CircleCI takes very long - so here's a PR using GH actions only.

It simplifies our docs build too, adds link checking and fixes a couple of links.


dholbach:
This is not urgent, but it'd reduce build / test times quite a lot. 1 minute compared to about 10 last I checked.

dholbach:
We'd need to turn off the Circle CI settings for grafanalib.

bboreham:
I'm not going to review this; I know nothing about GitHub actions.

PR #241, opened 2020-05-05
stop checking http://sphinx-doc.org/

dholbach:
stop checking http://sphinx-doc.org/ - the link is part of the theme and it times out right now


dholbach:
Merging this one too to test CI.

PR #249, opened 2020-05-14
Add Elasticsearch bucket script aggregator

DWalker487:
## What does this do?

This adds support for the Elasticsearch bucket script pipeline aggregator
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-bucket-script-aggregation.html 

## Why is it a good idea?

This is one of the core elasticsearch pipeline aggregations, and is natively supported by Grafana. 
It allows the user to combine the output of previous metric aggregations, using 
Elasticsearch Painless inline scripting.

## Context

Being able to apply scripts to the results of Elasticsearch metric aggregations can be very useful,
e.g. allowing one to plot the ratio of two sums over different document fields, as was my use case.

This includes id and hiding capabilities for the bucket script aggregator, as implemented elsewhere in #231

The Painless scripting docs are here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-painless.html

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Changelog
 Changes
 -------
 
+* Add Elasticsearch bucket script pipeline aggregator
 * ...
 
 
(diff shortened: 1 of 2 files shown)


matthewmrichter:
I don't have a means of testing this really but seems harmless enough. Please patch if you run into any bugs.

PR #258, opened 2020-07-01
Add support for the Stat panel, which replaces SingleStat

### LINKED ISSUE #253: Singlestat panel → Stat panel
The Singlestat panel has been deprecated and replaced with the [Stat ](https://grafana.com/docs/grafana/latest/panels/visualizations/stat-panel/)panel, so the Singlestat content has been removed. I will add some aliases to the Stat topic in case other people still have links to Singlestat.

_Originally posted by @oddlittlebird in https://github.com/weaveworks/grafanalib/issues/251#issuecomment-634645460_

JamesGibo:
## What does this do?
The SingleStat panel was deprecated in Grafana 7.0, Stat is thethe replacement for this.
The Stat panel uses a new mappings format so the classes StatMapping, StatValueMapping and StatRangeMapping were also added to support this.

## Why is it a good idea?
Allows users to make use of the Stat panel in the latest version of Grafana

## Context
Closes #253 

## Questions
The new mappings format is valid for not just the Stat panel, but didn't want to complicate this with the existing `Mapping` and `ValueMap` and `RangeMap` classes. Let me know if you think there is a better way to maintain backwards compatibility.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changes
 * Add Elasticsearch bucket script pipeline aggregator
 * Added ability to hide metrics for Elasticsearch MetricAggs
 * Add derivative metric aggregation for Elasticsearch
+* Add ``Stat`` class (and ``StatMapping``, ``StatValueMapping``, ``StatRangeMapping``) to support the Stat panel
 * ...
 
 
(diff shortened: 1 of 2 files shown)


Valiev:
Looking forward to see it merged in!

PR #263, opened 2020-09-28
Add transparency setting to Heatmap, Stat, PieChart panels

gjulianm:
## What does this do?

Adds a 'transparent' setting to three panels that missed it (Heatmap, Piechart, Stat).

## Why is it a good idea?

It was a missing setting, inconsistent with the rest of panels.

## Context

Nothing worthy of mention.

## Questions

I didn't put the change for 'Stat' in the CHANGELOG.rst file as it is a new panel for this version.


JamesGibo (approved):
LGTM

PR #264, opened 2020-10-09
A few blank lines as requested by the linter.

matthewmrichter:
## What does this do?
The pipeline's linter was complaining about formatting because we lost a few blank lines.


dholbach (approved):
:medal_sports:

PR #275, opened 2020-11-02
Make quote coding style consistent

JamesGibo:
## What does this do?
No functional change, have made quote style consistent across all files.

Triple Double quotes `"""` for docstrings
Double quotes `""` for human readable message or when string used for interpolation
Single quotes `''` for symbol like strings

## Why is it a good idea?
Makes it easier for contributors to know the correct coding style to use as it is consistent throughout all files

### DIFF
--- a/docs/CONTRIBUTING.rst
+++ b/docs/CONTRIBUTING.rst
@@ -33,6 +33,9 @@ Conventions
 * Local variables are ``snake_cased``
 * We're kind of fussy about indentation:
   4 spaces everywhere, follow the examples in `core.py`_ if you're uncertain
+* Triple Double quotes `"""` for docstrings
+* Double quotes "" for human readable message or when string used for interpolation
+* Single quotes '' for symbol like strings
 
 Testing
 -------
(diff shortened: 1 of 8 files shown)


dholbach (approved):
LGTM - I hadn't seen the `""` vs `''` distinction spelled out like this before. Is this very common? In general I like it and think it'll add more clarity.

JamesGibo:
There are a number of references to similar, such as these
https://luminousmen.com/post/the-ultimate-python-style-guidelines
https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python

I like it because it looks cleaner and clearer than just using single or double quotes everywhere
```Python
filter = attr.ib(default="")
# Better than
filter = attr.ib(default='')
```

```Python
obj = {
    'expr': self.expr
}
# Better than
obj = {
    "expr": self.expr
}
```

```Python
"""Function docstring

    xxxxxxx
    yyyyyyy
    """
# Better than
'''Function docstring

    xxxxxxx
    yyyyyyy
    '''
```
 
```Python
tmp = "rgba({}, {}, {}, {})".format(self.r, self.g, self.b, self.a)
# Better than
tmp = 'rgba({}, {}, {}, {})'.format(self.r, self.g, self.b, self.a)
```

dholbach:
Super. Thanks for the pointers and for looking into this!

PR #278, opened 2020-11-05
Adding Alert Threshold flag to Graphs.

mazzi:
## What does this do?
We found out that on [Grafana 7.2.0-beta1](https://github.com/grafana/grafana/blob/master/CHANGELOG.md#720-beta1-2020-09-09) adds toggle to disable alert threshold visibility in graph panel was added. I would like `grafanalib` to support it.

Selection_158

This PR adds an option on the Graph class to be able to toggle that setting (`alertThreshold`).

## Why is it a good idea?

To support this new flag on `grafanalib`.

## Context
Please check first title (What does this do). After checking how the json looked like on an instance of a graph using Grafana 7.3.0, decided to add the parameter that this PR describes to support the new attribute.
Didn't test on a real case. Just added the parameter and the serialization of the attribute.
Tests are passing locally.

## Questions
No questions from my side.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,8 @@ Changelog
 0.5.9 (TBD)
 ===========
 
+* Added Alert Threshold enabled/disabled to Graphs.
+
 Changes
 -------
 
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1061,6 +1061,7 @@ class Graph(object):
 
     title = attr.ib()
     targets = attr.ib()
+    alertThreshold = attr.ib(default=True, validator=instance_of(bool))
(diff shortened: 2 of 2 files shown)


JamesGibo (changes requested):
Looks good to me. 
Not tested due to being on an older version but have checked the
source code (https://github.com/grafana/grafana/blob/e269445d79ff4db740b8955a7251bbeb3d1e8775/public/app/plugins/panel/graph/module.ts#L142)

Please could you add a note to the Changelog?

mazzi:
> Please could you add a note to the Changelog?

Done ✅ 
Thanks @JamesGibo !

JamesGibo (approved):

PR #282, opened 2020-11-17
Add constants for all Grafana value formats

JamesGibo:
## What does this do?
Adds constants for all the Grafana value formats (https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/valueFormats/categories.ts)

## Why is it a good idea?
Easy to template the correct formatting required


dholbach (approved):
LGTM!

PR #284, opened 2020-11-20
Add support for repetition to the Stat panel

birdflyer-lszo:
Extend the Stat panel with support for repetitions.

I am using Stat panels in overview dashboards. Stat panels should be repeated according to the content of a variable (in my case the variable contains all tags from a certain measurement).

Without this little change, I had to generate the dashboard and either amend the missing repetitions by jq or later add them manually in Grafana. My intention is to generate the entire Grafana content completely automated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Changelog
 
 * Added Alert Threshold enabled/disabled to Graphs.
 * Added constants for all Grafana value formats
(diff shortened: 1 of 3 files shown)


JamesGibo (changes requested):
Looks goods, thanks for contributing.
Please could you add a note to the CHANGELOG about the functionality you have added?

JamesGibo:
Reference to feature: https://grafana.com/docs/grafana/latest/variables/repeat-panels-or-rows/

JamesGibo (approved):

PR #285, opened 2020-11-23
Add support for alias expressions in data targets

birdflyer-lszo:
Grafana allows setting an alias expression in a data target so that it gets identified nicely in the resulting graph. `Target` did not yet allow setting such an alias.

Add a simple `alias` field to the `Target` class so that the selected data can be aliased with an arbitrary expression.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Changelog
 * Added Alert Threshold enabled/disabled to Graphs.
 * Added constants for all Grafana value formats
 * Added support for repetitions to Stat Panels (https://grafana.com/docs/grafana/latest/variables/repeat-panels-or-rows/)
+* Added support for an alias expression in targets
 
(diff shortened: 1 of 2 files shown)


JamesGibo:
I wondered if it would be better to combine `alias` and `legendFormat` as they serve the same purpose, just for different data sources?
Or just include some docstrings for them?
See this for info on the difference for different data sources (https://community.grafana.com/t/how-to-override-metric-name/4388)

birdflyer-lszo:
You are right. If they are the same, there is little need for a separate property. I will kill this PR (I am using a different `Target` now anyway).

Will likely open one or the other PR in the future as I stumble across things i find missing :) The library is very helpful by the way!

JaSei:
Please do this. I use SqlTarget for influx (to be able to `hide`) and `legendFormat` doesn't work as alias

PR #287, opened 2020-12-02
add hide parameter to Target

eugenekainara:
## What does this do?

Add hide parameter to Target

## Why is it a good idea?

It's useful for cases where panel has few queries - one for displaying data and another one with hide=True - is used for Alerts


JamesGibo (changes requested):
Looks good, please could you add a note to the changelog?

JamesGibo (approved):

PR #289, opened 2020-12-10
Patch 1

shi-ron:
## What does this do?
Add textMode option in Stat Panel

## Why is it a good idea?
this option is missing


JamesGibo (approved):
Looks good, thanks for your contribution.
If you wanted, you could add some constants for the valid textMode key values, but not required.

shi-ron:
> Looks good, thanks for your contribution.
> If you wanted, you could add some constants for the valid textMode key values, but not required.

It's not necessary analogue to other classes ;)

PR #177, opened 2019-09-13
add all query settings to all query panels

butlerx:
## What does this do?

ensures all query panels have the same default query settings

## Why is it a good idea?

allows settings to be copied and defaults reused

PR #290, opened 2020-12-16
Create panel object for all panels to inherit 

JamesGibo:
## What does this do?
Ensures all query panels have the same default query settings

Based on PR #177

## Why is it a good idea?
Makes all panels consistent


dholbach (approved):
Thanks a lot for your work on this @JamesGibo and @butlerx!

PR #309, opened 2021-02-03
Enable use with pre-17.4.0 versions of attr.

banburybill:
PR #180 updated Grafanalib to work with attr versions post 19.2.0, when
keyword argument convert to attr.ib() was removed. It had been
deprecated in 17.4.0, and renamed to converter.

This change revises the change to work with both varieties of attr.ib(),
and so removes version requirements for attr. Those working on
antediluvian systems can use whatever system packaged version of attr is
available, a convenience when packaging Grafanalib.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         'Topic :: System :: Monitoring',
     ],
     install_requires=[
-        'attrs==20.3.0',
+        'attrs',
     ],
     extras_require={
         'dev': [
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -542,7 +542,10 @@ def _balance_panels(panels):
 class Row(object):
     # TODO: jml would like to separate the balancing behaviour from this
     # layer.
-    panels = attr.ib(default=attr.Factory(list), converter=_balance_panels)
+    try:
+        panels = attr.ib(default=attr.Factory(list), converter=_balance_panels)
+    except TypeError:
+        panels = attr.ib(default=attr.Factory(list), convert=_balance_panels)
     collapse = attr.ib(
         default=False, validator=instance_of(bool),
     )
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@ def local_file(name):
         'Topic :: System :: Monitoring',
     ],
     install_requires=[
-        'attrs==20.3.0',
```
JamesGibo, line 43: Could we instead set a minimum supported version?

JamesGibo:
Hi Jim,
Thanks for your PR, the code looks good.

I just have a questions and a word of warning I guess. 
What requirement do you have for supporting versions of attrs before version 17.4.0, which was released in Dec 2017?
https://www.attrs.org/en/stable/api.html#attr.ib
I just wondered, as removing the minimum required version of attrs from the requirements might have other unintended compatibility issue.

@dholbach what is your opinion?

banburybill:
My specific use case here is with a production system on a worldwide install of ~250 hosts which is still running Ubuntu Xenial. Xenial remains in support until April this year. Having said that, I am currently including a private install of attr in the source tree for the application in question, and the shipping product includes Grafanalib product not Grafanalib itself, so while being able to just use the system packaged version would be nice (we build on Xenial for obvious reasons), it's not a big deal for me. Once they move to a later LTS, all will be happy again. This just seemed a small change to the original PR, and it might be worth a try :-)

setup.py
banburybill: For me, setting a minimum version rather than requiring a specified version would be better. Specifying an individual version (even if kept up to date) does suggest some sort of dark reliance on internal implementation details in attr.

JamesGibo: What minimum version do you require? Please could you add this as the minimum in your PR?

JamesGibo (changes requested):
I am happy to switch to a minimum version requirement for attrs

setup.py
banburybill: For my particular use case, 15.2.0 would be fine, and a version I have tested.

setup.py
```diff
@@ -40,7 +40,7 @@ def local_file(name):
...
+        'attrs',
```
JamesGibo, line 43:
```suggestion
        'attrs>=15.2.0',
```

JamesGibo:
Closing see #321 for more info

PR #317, opened 2021-02-25
add Notification class

olha-honcharuk:
## What does this do?

Add `Notification` class
## Why is it a good idea?

It will allow to set up and configure `notifications` on `Alert`


JamesGibo (approved):
Looks good

PR #326, opened 2021-03-09
Add support for thresholds to graph panels

### LINKED ISSUE #171: Feature: support "thresholds" in Graph
## What you expected to happen?
Graph doesn't support [thresholds](https://grafana.com/docs/features/panels/graph/#thresholds)

## What happened?

I wanted to specify a threshold for a graph, but it seems [Graph](https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py) class doesn't support it. 

## How to reproduce it?

N/A

This seems to be a minor change, so unless there is ongoing work to add this I'll try to send PR soon.

JamesGibo:
## What does this do?
Add support for thresholds to graph panels

This code has not been tested if the thresholds created work in practice, but follows the same pattern as other panels

Closes #171

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1332,6 +1332,7 @@ class Graph(Panel):
         if self.alert:
             graphObject['alert'] = self.alert
             graphObject['alertRuleTags'] = self.alertRuleTags
+            graphObject['thresholds'] = []
         if self.thresholds and self.alert:
             print("Warning: Graph threshold ignored as Alerts defined")
         return self.panel_json(graphObject)
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1171,13 +1171,17 @@ class Graph(Panel):
     """
     Generates Graph panel json structure.
 
-    :param dataLinks: list of data links hooked to datapoints on the graph
+    :param alert: List of AlertConditions
+    :param alertRuleTags: Key Value pairs to be sent with Alert notifications
+    :param dataLinks: List of data links hooked to datapoints on the graph
     :param dataSource: DataSource's name
     :param minSpan: Minimum width for each panel
     :param repeat: Template's name to repeat Graph on
-    :param thresholds: graph panel thresholds
+    :param thresholds: List of GraphThresholds - Only valid when alert not defined
     """
 
+    alert = attr.ib(default=None)
+    alertRuleTags = attr.ib(default=attr.Factory(dict))
     alertThreshold = attr.ib(default=True, validator=instance_of(bool))
     aliasColors = attr.ib(default=attr.Factory(dict))
     bars = attr.ib(default=False, validator=instance_of(bool))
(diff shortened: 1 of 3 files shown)


grafanalib/core.py
```diff
@@ -1238,6 +1240,7 @@ def to_json_data(self):
             'stack': self.stack,
             'steppedLine': self.steppedLine,
             'tooltip': self.tooltip,
+            'thresholds': self.thresholds,
```
olha-honcharuk, line 1243: I believe that `thresholds` is configurable only if there is no alert set, otherwise `thresholds` will be simply ignored.

grafanalib/core.py
```diff
@@ -2585,11 +2592,20 @@ def to_json_data(self):
 
 @attr.s
 class Threshold(object):
-    """Threshold for a gauge
+    """Threshold for for panels
+    (https://grafana.com/docs/grafana/latest/panels/thresholds/)
```
olha-honcharuk, line 2596: I suppose that the JSON model for panel threshold is a bit different. It should contain `colorMode` with `ok`,  `warning`, `critical`, `custom` values; boolean `fill`; not null `value` and optional `fillColor` and `lineColor` for `custom` mode if the corresponding option is enabled.

JamesGibo: Thanks for the review, I will look at this towards the end of the week.

JamesGibo: I have added a separate class for Graph panels to support this

grafanalib/core.py
JamesGibo: I have added a printed waning and something to the docs

olha-honcharuk: It should produce a valid config for the dashboard, `thresholds` will be just ignored. So I think a warning is enough.

olha-honcharuk: Also, we can define it in `else` of `if self.alert` block simply ignoring when we have an alert defined.

grafanalib/core.py
```diff
@@ -2692,6 +2713,50 @@ def to_json_data(self):
...
+    """
+
+    value = attr.ib(validator=instance_of(float))
+    colorMode = attr.ib(default="critical")
+    fill = attr.ib(default=True, validator=instance_of(bool))
+    line = attr.ib(default=True, validator=instance_of(bool))
+    op = attr.ib(default=EVAL_GT)
+    yaxis = attr.ib(default='left')
+    fillColor = attr.ib(default=RED)
+    lineColor = attr.ib(default=RED)
```
olha-honcharuk, line 2745: `fillColor` and `lineColor` should be defined only for `custom` `colorMode`. I don't think a default value for them should be defined as they are optional.

olha-honcharuk (approved):
LGTM, thanks!

grafanalib/core.py
JamesGibo: I have made them conditional, but have left the defaults as didnt want to just put empty string

PR #328, opened 2021-03-12
Adding humio datasource. 

jeremiah1066:
Humio is available as a plugin datasource, documented [here](https://grafana.com/grafana/plugins/humio-datasource/).
The Humio query language is documented [here](https://docs.humio.com/reference/language-syntax/). 

## What does this do?
Adding support for Humio queries in grafanalib.

## Why is it a good idea?
Humio is offered as a plugin datasource in Grafana, and adding this will allow people to create graphs with Humio data. 

## Context
Humio is a log aggregation, storage, and analysis tool similar to Elasticsearch. It is available as both as self-hosted or as a SaaS product.

### DIFF
--- a/grafanalib/tests/test_humio.py
+++ b/grafanalib/tests/test_humio.py
@@ -6,7 +6,7 @@ from grafanalib import _gen
 from io import StringIO
 
 
-def test_serialization_cloudwatch_metrics_target():
+def test_serialization_humio_metrics_target():
     """Serializing a graph doesn't explode."""
     graph = G.Graph(
         title="Humio Logs",
--- a/grafanalib/tests/test_humio.py
+++ b/grafanalib/tests/test_humio.py
@@ -1,0 +1,25 @@
+"""Tests for Humio Datasource"""
+
+import grafanalib.core as G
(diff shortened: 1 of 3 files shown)


grafanalib/tests/test_humio.py
```diff
@@ -0,0 +1,25 @@
+"""Tests for Humio Datasource"""
+
+import grafanalib.core as G
+import grafanalib.humio as H
+from grafanalib import _gen
+from io import StringIO
+
+
+def test_serialization_cloudwatch_metrics_target():
```
JamesGibo, line 9:
```suggestion
def test_serialization_humio_metrics_target():
```

JamesGibo (changes requested):
Thanks for the PR, its great to see support for new data sources. 

The PR looks good just one minor change to function name requested

grafanalib/tests/test_humio.py
jeremiah1066: @JamesGibo Oops, good catch. Just updated the PR and cleaned up some of the comments too.

JamesGibo (approved):

PR #331, opened 2021-03-18
Fix Repeat for Panel

charlesoconor:
## What does this do?

Repeat is an object so isn't json serializable this change will add the necessary fields to the pannel object so that we can create repeat panels. 

## Why is it a good idea?

This is a bug fix. 

## Context

Tested locally and was able to create repeated panels.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1263,7 +1263,6 @@ class Graph(Panel):
     :param dataLinks: List of data links hooked to datapoints on the graph
     :param dataSource: DataSource's name
     :param minSpan: Minimum width for each panel
-    :param repeat: Template's name to repeat Graph on
     :param thresholds: List of GraphThresholds - Only valid when alert not defined
     """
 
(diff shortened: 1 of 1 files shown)


JamesGibo (changes requested):
Thanks for spotting this and submitting a fix. 

Please could you also remove the reference to repeat from a couple of places?
https://github.com/weaveworks/grafanalib/blob/36ee609a99f9638c95c9c8be92532a8843e8cd5a/grafanalib/core.py#L1534
https://github.com/weaveworks/grafanalib/blob/36ee609a99f9638c95c9c8be92532a8843e8cd5a/grafanalib/core.py#L1548
https://github.com/weaveworks/grafanalib/blob/36ee609a99f9638c95c9c8be92532a8843e8cd5a/grafanalib/core.py#L1266

charlesoconor:
Updated to remove the old refs to repeated in things that have subclasses `Panel`.  I left the repeated value on the `Row` class since it's marked legacy for the old Grafana schema. Can update/remove that if it makes sense? 

https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L1227

JamesGibo (approved):
Looks great, thanks for your contribution.

PR #334, opened 2021-03-24
Add timeField field for the Elasticsearch target to allow the alert t…

dafna-starkware:
## What does this do?
Add timeField field for the Elasticsearch target 

## Why is it a good idea?
It able the alert to change its state.


JamesGibo (changes requested):
Thanks for the PR which looks great to me.

Please could you move the line in the change log up to the unreleased version as version 0.5.10 has now been released?

JamesGibo (approved):

PR #346, opened 2021-04-13
Table driven example with simple upload script

rcowham:
## What does this do?
Shows simple alternative example of generating a dashboard from static table of metrics.

## Why is it a good idea?
Helps people get started faster

## Questions
Unsure about CHANGELOG format

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Changes
 * Changed ...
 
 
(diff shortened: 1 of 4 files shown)


JamesGibo (approved):
Thanks for this contribution, hopefully it will be of use to people getting started with grafana and grafanalib.
When we get round to updating the docs further will make sure to include some greater sign posting to this example.

Issue #340, opened 2021-03-31
Grafana v7.5.1 and possibly earlier require datasource to be set in target


JamesGibo:
Using Grafana v7.5.1 and possibly earlier versions if you do not set the datasource in your target, the query will not be populated.

To fix set the datasource in the target
```python
Stat(
        title="My stat",
        description="""xxxxx""",
        dataSource='prometheus',
        targets=[
            Target(
                expr='sum(irate(nginx_http_requests_total{job="default/frontend",status=~"1.."}[1m]))',
                legendFormat="1xx",
                refId='A',
                datasource='prometheus'
            ),
        ],
        gridPos=GridPos(h=5, w=4, x=0, y=1)
)
```

Need to change the example in the docs to cover this and possibly add a warning in the docs after finding which version of grafana this was introduced in

ttk:
I found that the generated JSON produces `"datasource": "",` which doesn't work...however, if set to `"datasource": null,` then it works correctly.

JamesGibo:
> I found that the generated JSON produces `"datasource": "",` which doesn't work...however, if set to `"datasource": null,` then it works correctly.

Was that for the datasource in the target? Or the one for the panel? 
```
Target(
                expr='sum(irate(nginx_http_requests_total{job="default/frontend",status=~"1.."}[1m]))',
                legendFormat="1xx",
                refId='A',
                datasource='prometheus'
            ),
```

ttk:
In my case, this was for the `Target` where I didn't specify a `datasource` in the python code.  Looks like the default `datasource` is set to empty string `""` but this apparently doesn't work recent versions of Grafana.  

In `class Target` if I change `datasource = attr.ib(default="")` to `datasource = attr.ib(default=None)` then it works as expected.

JamesGibo:
That seems to match the pattern used for a panel datasource.  
Would you mind submitting a PR for that change?

https://github.com/weaveworks/grafanalib/blob/83c42c1e7661da1f06c1792bbe300c112d15791a/grafanalib/core.py#L1128

JamesGibo:
@dgzlopes Hopefully this PR will fix you issue

dgzlopes:
Nice! Thanks for the poke @JamesGibo.

I'll try :)

PR #348, opened 2021-04-15
Set target datasource as None instead of empty string

ttk:
## What does this do?
Changes the default for target datasource

## Why is it a good idea?
The generated dashboard now works with newer versions of Grafana.

See:
https://github.com/weaveworks/grafanalib/issues/340


JamesGibo:
Closes #340

JamesGibo:
Thanks for contribution

PR #349, opened 2021-04-16
Remove re-defined maxDataPoints from multiple panels

dgzlopes:
## What does this do?
This PR removes the maxDataPoints field from multiple panels b/c it's inherited from Panel.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,10 @@ x.x.x (TBD)
 ==================
 
 * Added ...
+* Added timeField field for the Elasticsearch target to allow the alert to change its state
+* Added nameFilter field for the AlertList panel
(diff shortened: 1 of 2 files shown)


JamesGibo:
Thanks for the PR, just had a quick look and this should present in the panel already as it is defined in the Panel object inherited by all panels. 
https://github.com/weaveworks/grafanalib/blob/e965fe916572934eaef55e9859a4bd3f10ef06e1/grafanalib/core.py#L1141

I see that some panels have it re-defined, so maybe it should be removed from them instead?

dgzlopes:
Didn't see that! Yup, makes a lot of sense. Changed :)

JamesGibo (approved):
Looks great, thanks for tidying up this attribute

PR #354, opened 2021-04-19
Fix name filter field.

dafna-starkware:
What does this do?
Fix a small bug in the nameFilter field of the AlertList panel, add a test for the class and expand its' documentation.


JamesGibo (approved):
LGTM, thanks for adding to the unit tests and documentation

PR #355, opened 2021-04-23
Added 'hide' parameter to CloudwatchMetricsTarget class

bnogas:
## What does this do?
This PR adds 'hide' parameter to CloudwatchMetricsTarget class

## Why is it a good idea?
hide parameter is useful when you need to use a metric, but you don't want to display it. 

Note that 'hide' parameter is already defined for the `Target` class.

## Context
Example:
```
 Graph(
           # (...)
            targets=[
                CloudwatchMetricsTarget(
                    # (...)
                    metricName="SignInSuccesses",
                    namespace="AWS/Cognito",
                    alias="OK",
                    id="ok"
                ),
                CloudwatchMetricsTarget(
                    # (...)
                    metricName="SignInSuccesses",
                    namespace="AWS/Cognito",
                    id="total",
                    hide=True,
                ),
                CloudwatchMetricsTarget(
                    # (...)
                    id="failed",
                    expression="SUM([total, -ok])"
                ),
            ],
        ),
    ],
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Changes
 * Added table-driven example dashboard and upload script
 * Removed re-defined maxDataPoints field from multiple panels
 * Fix the AlertList class and add a test for it
+* Added hide parameter to CloudwatchMetricsTarget class
 
 
 0.5.11 (2021-04-06)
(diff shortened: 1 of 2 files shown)


JamesGibo:
Thanks for the PR, looks great. 
I just added a validator to the attribute

JamesGibo (approved):

bnogas:
Thank you for the instant merge :) !

PR #359, opened 2021-04-26
Change 'target' attribute validator to a generic one in 'AlertCondition'

bnogas:
* Accept 'CloudwatchMetricsTarget' as target for 'AlertCondition'

## What does this do?
This PR changes 'target' attribute validator in AlertNotifications class

## Why is it a good idea?
I've encountered an issue where 'AlertNotifcation' class didn't accept CloudwatchMetricsTarget as target

## Questions
I've been considering creation of zope.interface.Interface and veryfing based on attrs.validators.provides, but this simple solution seems to be good enoguh for me

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ x.x.x (TBD)
 Changes
 -------
 
-*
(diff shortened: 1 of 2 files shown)


JamesGibo (approved):
LGTM, sure we could make the validation more strict but don't think that is really necessary

PR #360, opened 2021-04-28
Move alertRuleTag from Graph to Alert.

dafna-starkware:
## What does this do?
Move the alertRuleTags field from Graph to Alert.

## Why is it a good idea?
Because the dashboard extracts the alertRuleTags field from an alert mapping.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/grafanalib/360)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,12 +5,13 @@ Changelog
 x.x.x (TBD)
 ==================
 
-* Added ...
+* Added a test for the Alert class.
 
 Changes
 -------
 
 * Bugfix: changed 'target' validator in AlertNotification to accept CloudwatchMetricsTarget
+* Moved the alertRuleTag field from Graph to Alert.
 
 0.5.12 (2021-04-24)
 ==================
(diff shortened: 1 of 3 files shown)


JamesGibo (commented):
LGTM, thank the for PR.
Have you found any documentation about the way the dashboard extracts the alertRuleTags field from an alert mapping?
I am a bit hesitant to remove it from the Graph panel in case someone is using it, but if it never work then would be happy to remove it?

dafna-starkware:
Hi James, 
Thanks for the review!
In [this](https://grafana.com/docs/grafana/latest/alerting/create-alerts/) documentation you can see that the tag is part of the alert configuration. I tried to use the current code to add an alert rule tag but it didn't work. When I did it when the AlertRuleTag field was part of the Alert field it did work.

bnogas:
I just hit the same issue. I confirm that alertTagsRules are currently mislocated in the resulting json.

Here is an example from [API documentation](https://grafana.com/docs/grafana/latest/http_api/dashboard/) (serach for alertRuleTags)
Here is the current implementation in [grafana](https://github.com/grafana/grafana/commit/e06abb30aa7f234bcc906bcbdee2ef3936b0d476#diff-a5b1d2b0601f1005bd1cada282898fa009b183db3af8fbbe34d4d367b0c13666)

JamesGibo (approved):

PR #362, opened 2021-04-30
Fix: 'alertRuleTags' are now defined in 'Alert' class instead of 'Graph'

bnogas:
## What does this do?
This PR moves 'alertRuleTags' from 'Graph' to 'Alert' class

## Why is it a good idea?
At least in Grafana version  v.7.5.2 alert is the right place to define alertRuleTags


bnogas:
it's a duplicate  https://github.com/weaveworks/grafanalib/pull/360

PR #367, opened 2021-05-17
Add transformations field for the Panel class

strangeman:
## What does this do?
It's pretty straightforward code, which allows to add transformations (https://grafana.com/docs/grafana/next/panels/transformations/types-options/#transformation-types-and-options) for the Table panel. At the current moment, it just translate list of dicts with transformations to JSON as-is.

## Why is it a good idea?
I really miss this feature for the Tables. Adding transformations to the tables is very common pattern, which allows to increase readability and usability of Tables.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1142,6 +1142,7 @@ class Panel(object):
     :param timeFrom: time range that Override relative time
     :param title: of the panel
     :param transparent: defines if panel should be transparent
+    :param transformations: defines transformations applied to the table
     """
 
     dataSource = attr.ib(default=None)
@@ -1164,6 +1165,7 @@ class Panel(object):
     timeFrom = attr.ib(default=None)
     timeShift = attr.ib(default=None)
     transparent = attr.ib(default=False, validator=instance_of(bool))
+    transformations = attr.ib(default=attr.Factory(list), validator=instance_of(list))
 
     def _map_panels(self, f):
         return f(self)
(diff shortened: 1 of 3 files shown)


grafanalib/core.py
```diff
@@ -2033,6 +2034,7 @@ class Table(Panel):
     sort = attr.ib(
         default=attr.Factory(ColumnSort), validator=instance_of(ColumnSort))
     styles = attr.ib()
+    transformations = attr.ib(default=list(), validator=instance_of(list))
```
JamesGibo, line 2037:
```suggestion
    transformations = attr.ib(default=attr.Factory(list), validator=instance_of(list))
```

JamesGibo (commented):
Thanks for the PR looks good, please could apply the suggested change?
Optional: if you have chance please could you add a simple test for using a transform.

strangeman:
Thank you for review! Yeah, I'll add a test.

strangeman:
> Optional: if you have chance please could you add a simple test for using a transform.

Done

JamesGibo:
Thanks,
I also spotted that this is relevant to all panels from the link you shared, so should in be add to the Panels class instead?
https://github.com/weaveworks/grafanalib/blob/b581a6042f5294125dd72b7bf0aff4cdbbb087e3/grafanalib/core.py#L1123

strangeman:
> Thanks,
> I also spotted that this is relevant to all panels from the link you shared, so should in be add to the Panels class instead?
> 
> https://github.com/weaveworks/grafanalib/blob/b581a6042f5294125dd72b7bf0aff4cdbbb087e3/grafanalib/core.py#L1123

Sure, done! Fixed flake8 errors also

JamesGibo (approved):

JamesGibo:
Thanks for you contribution I have now merged this.

At some point we should add a Transformation class for this #367

PR #370, opened 2021-05-19
Add documentation about generating dashboards from code

strangeman:
I experienced lack of documentation when tried to generate dashboard directly from code, without `generate-dashboard` tool, so I added code sample for this case.


JamesGibo (approved):
Thanks, always good to add some examples. 

Probably need to add make this more visible in the docs aswell (https://github.com/weaveworks/grafanalib/blob/master/grafanalib/tests/examples/upload_grafana_dashboard.sh)

PR #371, opened 2021-05-21
Remove Bryan Boreham as maintainer

bboreham:
I am moving on from Weaveworks.


JamesGibo:
Good look in you new role, thanks for your contributions to grafanalib. 

@dholbach will the write access be handled automatically by weaveworks or is that something you need to change?

dholbach:
It should be handled by Weaveworks IT.

dholbach (approved):

Issue #379, opened 2021-06-14
Add support for fillGradient


mr-karan:
`fillGradient` property was added in [Grafana 6.3.](https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v6-3/#graph-gradients)

Currently this package only support `fill`. Maybe we can add `fillGradient` too.

I believe it needs to be added here; https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L1294

Issue #124, opened 2018-05-09
XAxis Object missing Attributes


spprichard:
## What you expected to happen?
When creating an `XAxis` Object expect more then just the `show` attribute on the XAxis Object to be translated into JSON. 
## What happened?

There is no error message in this case. However, if you look at the JSON that is generated, only the `show` key is set. 
```json
"xaxis": {
    "show": true
}, 
```

I expect the following:
```json
"xaxis": {
    "mode": "series", 
     "name": null, 
     "show": true, 
     "values": [
         "current"
     ]
}, 
```

## How to reproduce it?

- Create a Graph Object
- Give it a `xAxis` attribute 
- Create a `XAxis` _see below_
```json
xAxis=XAxis(
    show=True,
    mode="series",
    values=["current"],
),
```
- Observe generated JSON
```json
"xaxis": {
     "show": true
}, 
```

I would propose that the XAxis Class should look like this:
```python
@attr.s
class XAxis(object):

    mode = attr.ib(default="time", validator=is_valid_xaxis_mode)
    name = attr.ib(default=None)
    values = attr.ib(default=attr.Factory(list))
    show = attr.ib(validator=instance_of(bool), default=True)

    def to_json_data(self):
        return {
            'show': self.show,
            'mode': self.mode,
            'values': self.values,
            'name' : self.name,
        }
```

ushasm:
Even I am face the same issue. 
Can we commit the changes and create PR for the same ?

jml:
PRs are very welcome!

ushasm:
Seems like we do not have permission to create a feature branch and commit the changes :(

3v1lm1ky:
I have same issue right here

JamesGibo:
Closed in #369

PR #380, opened 2021-06-30
Add extraJson attribute to the Panel class for overriding the panel with raw JSON

Tenchi2xh:
## What does this do?

This adds a new attribute called `extraJson` to the `Panel` class, enabling users to add attributes that are not currently supported by grafanalib, for example `fillGradient` and color mappings:

Example 1, a Graph with aligned y-axes and a fill gradient:

```py
Graph(
    title="...",
    dataSource=datasource,
    yAxes=YAxes(
        left=YAxis(format="bytes", min=None),
        right=YAxis(format="none", min=None),
    ),
    targets=[
        # ...
    ],
    extraJson={
        "yaxis": {"align": True},
        "fillGradient": 6
    }
)
```

Example 2, a pie chart with combining enabled and percentage shown in the legend:

```py
PieChart(
    title="..",
    dataSource=datasource,
    targets=[
        # ...
    ],
    pieType="donut",
    extraJson={
        "legend": {"percentage": True},
        "combine": {"label": "Others", "threshold": "0.01"}
    }
)
```

Other nice uses is for gradient color mappings:

- For `GaugePanel`:  

    ```py
        extraJson={"options": {"fieldOptions": {"defaults": {"color": {"mode": "continuous-GrYlRd"}}}}},
    ```
- For `Stat`: 

    ```py
        extraJson={"fieldConfig": {"defaults": {"color": {"mode": "continuous-GrYlRd"}}}},
    ```

Also supports adding options to nested objects, for example to a `Legend`. In the above example of a `PieChart`, if the user has already defined `legend=Legend(...)` and `extraJson` has an override for `legend`, the `Legend` object will first be converted using `to_json_data()` and then updated recursively.

## Why is it a good idea?

- Enables temporary support for Grafana features not yet supported by grafanalib
- Reduces users' boilerplate of manually patching the generated JSON output

Would temporarily help #379 #124 and possibly others

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ x.x.x (TBD)
 * Added missing attributes from xAxis class
 * Added transformations for the Panel class (https://grafana.com/docs/grafana/next/panels/transformations/types-options/#transformation-types-and-options)
 * Added Worldmap panel (https://grafana.com/grafana/plugins/grafana-worldmap-panel/)
+* Added ``extraJson`` attribute to the Panel class for overriding the panel with raw JSON
 
 Changes
 -------
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1131,6 +1131,20 @@ class Dashboard(object):
         }
 
 
+def _deep_update(base_dict, extra_dict):
+    if extra_dict is None:
+        return base_dict
+
+    for k, v in extra_dict.items():
+        if k in base_dict and hasattr(base_dict[k], "to_json_data"):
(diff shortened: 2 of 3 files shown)


JamesGibo (changes requested):
Hi, 
Thanks for the PR it looks good and will be a really useful feature, please could you add a docstring line to the Panel class describing  the new attribute you have added `extraJson`?

Hopefully users will still raise an issue for any missing features, so we can get them properly added for all users.

JamesGibo (approved):

Issue #110, opened 2018-02-16
Document the use of Templating


erkolson:
## What you expected to happen?
I attempted to use add templating to a dashboard, it is not documented how to use it so I provided these parameters:
https://github.com/aknuds1/grafanalib/blob/master/grafanalib/core.py#L431-L439
```python
        :param default: the default value for the variable
        :param dataSource: where to fetch the values for the variable from
        :param label: the variable's human label
        :param name: the variable's name
        :param query: the query users to fetch the valid values of the variable
        :param allValue: specify a custom all value with regex,
            globs or lucene syntax.
        :param includeAll: Add a special All option whose value includes
            all options.
```

## What happened?
Js error in grafana when trying to import the resulting json dashboard.  After some effort I determined that the js error is caued by the absence of some of these values for each template:
https://github.com/aknuds1/grafanalib/blob/master/grafanalib/core.py#L456-L470
```python
    def to_json_data(self):
        return {**_as_dict(self), **{
            'current': {
                'text': self.default,
                'value': self.default,
                'tags': [],
            },
            'hide': 0,
            'multi': False,
            'options': [],
            'refresh': 1,
            'regex': '',
            'sort': 1,
            'tagValuesQuery': None,
            'tagsQuery': None,
            'type': 'query',
        }}
```

I'm not very familiar with python but that bit of code looks like it should be used as defaults?  In any case, when I include those keys/values for each templated variable in my dashboard.py, the resulting json generated does work!

## How to reproduce it?
Create any dashboard that uses the templating, i.e.:
```python
  templating=Templating(list=[
    { # namespace
      "default": "",
      "datasource": "prometheus",
      "label": "Namespace",
      "name": "namespace",
      "query": "label_values(kube_pod_info, namespace)",
      "allValue": ".+",
      "includeAll": True,
    },
  ]),
```

Is this a bug with how templating gets rendered?  If not, can you add the proper incantation to use templating in the example dashboard.py?

jml:
Thanks! This looks like a bug. I hope to try to fix it soon, but am stretched pretty thin.

taion809:
@jml do you have any thoughts as to where this bug might be?  I looked through the code a bit and it's not really clear why this would work as expected.

jml:
I think it's just missing values in one of the objects. Isolating it is made more complicated by the fact that the bug is only revealed when the generated JSON is loaded by Grafana itself.

I won't have time to get into this for a little while.

[...]

PR #390, opened 2021-09-04
Templating

chrisfleming:
## What does this do?

This ensures that the selected value is set, for templating, I was finding that without setting selected the templating just didn't work when I loaded my dashboard.

## Why is it a good idea?

I think that this works to help resolve https://github.com/weaveworks/grafanalib/issues/110

In my case, I 'm using the following template, and it's working as I would expect:

` Template(
    dataSource="Health",
    name="Host", 
    refresh=2,
    query={"find": "terms", "field": "host.keyword"}, 
 ) `

## Questions

I've not had time to look in detail on the self.type == 'custom side' side of the if. But I can see that selected is getting set, so I don't think that case has this issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ x.x.x (TBD)
 * Added missing show percentage attribute to Pie chart panel
 * Added ``extraJson`` attribute to the Panel class for overriding the panel with raw JSON
 * Added inline script support for Elasticsearch metrics
+* Seletcted needs to be set as a bool value for templating to work.
 
 Changes
 -------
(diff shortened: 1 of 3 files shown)


JamesGibo (changes requested):
Looks good, i think this requirement must have been added in a newer version of grafana as I have been using Template and not had this issue up until now.
For this PR could you also add a line to the CHANGELOG?

JamesGibo (approved):

JamesGibo:
Thanks again for the PR, I will make a release shortly as there is quite a bit in the changelog now

PR #391, opened 2021-09-04
Add inline script support for Elasticsearch metrics

chrisfleming:
## What does this do?
This adds support for inline scripts to Elasticsearch metrics.

## Why is it a good idea?
Allows users to apply calculations to metrics directly

## Context
We have some metrics stored as Bytes/s but want to show them as bits/s, this patch allows the setting of inline to allows the calculation to be performed. 

With this patch, if inline is set then the output json looks something like 

`
 {  
   "field": "port_1_bytes",
   "hide": false,
   "id": "4", 
   "inlineScript": "_value*8",
   "meta": {}, 
   "settings": {  
        "script": { 
            "inline": "_value*8"
        }
     },
     "type": "max"
},        
`

If inline isn't set then we use:

`
 {  
   "field": "port_1_bytes",
   "hide": false,
   "id": "4", 
   "inlineScript": "_value*8",
   "meta": {}, 
   "settings": {},
     "type": "max"
},        
`

There is one additional change here, I found that for DateHistogramGroupBy, I had too add  'field': 'select field', to get this working.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ x.x.x (TBD)
 * Added missing align to graph panel
 * Added missing show percentage attribute to Pie chart panel
 * Added ``extraJson`` attribute to the Panel class for overriding the panel with raw JSON
+* Added inline script support for Elasticsearch metrics
 
 Changes
 -------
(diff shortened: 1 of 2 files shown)


JamesGibo (changes requested):
Thanks for you contribution, looks good to me but have not testing in practice.
Please could you fix the minor linting errors and add a note to the changelog?

JamesGibo:
Thanks for those changes for linting, if possible please could you add a line to the changelog as well?

JamesGibo (approved):

PR #399, opened 2021-09-22
Do not default to 0 for min value for yAxes

locan11:
Setting 0 as default min value for yAxes is arbitrary and messes up all of Grafana's auto-scaling features. This commit sets the default to None and lets Grafana choose the min value.


JamesGibo:
Thanks for your contribution!
PR looks good, please could you add a note to the change log under 'Changes'? 
Just so if people are already using 0 there are aware of the change.

JamesGibo:
Merged without changed will add in later commit

PR #403, opened 2021-09-29
Discrete panel

milkpirate:
## What does this do?
Adds basic support for Discrete https://grafana.com/grafana/plugins/natel-discrete-panel/ including panel itself and color mappings (DiscreteColorMappingItem to keep things tidy).

## Why is it a good idea?
Adds "Discrete".

## Context
Discrete inherits from Panel to keep the changes small.

## Questions
- How extensiv are you testing?
- Do these tests suffice?
- Any suggestions to use pre-existing features instead of DiscreteColorMappingItem?

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1452,6 +1452,7 @@ class Graph(Panel):
 
         def set_refid(t):
             return t if t.refId else attr.evolve(t, refId=next(auto_ref_ids))
+
         return self._map_targets(set_refid)
 
 
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1495,10 +1495,6 @@ class Discrete(Panel):
     def to_json_data(self):
         graphObject = {
             'colorMaps': self.colorMapsItems,
-            'datasource': self.dataSource,
-            'grid': self.gridPos,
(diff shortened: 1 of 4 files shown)


grafanalib/core.py
```diff
@@ -1452,9 +1453,57 @@ def auto_ref_ids(self):
...
+        ),
+    )
+
+    def to_json_data(self):
+        graphObject = {
+            'colorMaps': self.colorMapsItems,
+            'datasource': self.dataSource,
+            'grid': self.gridPos,
+            'targets': self.targets,
+            'title': self.title,
```
JamesGibo, line 1501: These lines can all be removed as they will be inherited from the Panel class
```suggestion
```

JamesGibo (changes requested):
Thanks for this PR, just one minor change to remove code not required

grafanalib/core.py
milkpirate: Accepted.

JamesGibo (approved):

JamesGibo:
Thanks for your contribution @milkpirate

milkpirate:
I might add functionality in the future for the many switches the panel supports:
image

PR #406, opened 2021-09-30
Add support for State Timeline and new Table panel

JamesGibo:
## What does this do?
Add support for State Timeline and new Table panel

## Questions
The only issue is that to support the Table panel going forward I have had to make breaking changes to the Table class, i could have just made a new class, but that starts to get messy. So thought it better to just make a breaking change and user that have not updated to v8 yet should pin to using an older version of grafanalib.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -2422,33 +2422,6 @@ class Column(object):
         }
 
 
-def _style_columns(columns):
-    """Generate a list of column styles given some styled columns.
-
-    The 'Table' object in Grafana separates column definitions from column
-    style definitions. However, when defining dashboards it can be very useful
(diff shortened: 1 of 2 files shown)


grafanalib/core.py
```diff
@@ -2196,78 +2201,59 @@ def _style_columns(columns):
...
-                pattern='time',
-                type=DateColumnStyleType(),
-            ),
-            ColumnStyle(
-                pattern='/.*/',
-            ),
-        ]
-
-    @classmethod
-    def with_styled_columns(cls, columns, styles=None, **kwargs):
```
dholbach, line 2242: Is is safe to remove this method?

PR #407, opened 2021-10-01
Discrete options

milkpirate:
## What does this do?
Adds additional options for Discrete panel.

## Why is it a good idea?
Makes the Dicrete panel more customizable.

## Note:
Had to shuffel some classes around. Git represents that quite ugly :/

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1637,6 +1637,7 @@ class Discrete(Panel):
     decimals = attr.ib(default=0, validator=instance_of(int))
     legendPercentDecimals = attr.ib(default=0, validator=instance_of(int))
     rowHeight = attr.ib(default=50, validator=instance_of(int))
+    textSize = attr.ib(default=24, validator=instance_of(int))
 
     units = attr.ib(default="none", validator=instance_of(str))
     legendSortBy = attr.ib(
@@ -1694,6 +1695,7 @@ class Discrete(Panel):
             'legendPercentDecimals': self.legendPercentDecimals,
             'decimals': self.decimals,
             'rowHeight': self.rowHeight,
+            'textSize': self.textSize,
 
             'units': self.units,
             'legendSortBy': self.legendSortBy,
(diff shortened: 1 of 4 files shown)


grafanalib/core.py
```diff
@@ -1570,24 +1553,172 @@ def to_json_data(self):
...
+            "color": self.color,
+            "text": self.text,
         }
 
 
+@attr.s
+class Discrete(Panel):
+    """
+    Generates Discrete panel json structure.
+
```
JamesGibo, line 1578: Add link to panel
```suggestion

https://grafana.com/grafana/plugins/natel-discrete-panel/
```

JamesGibo (approved):
Thanks for the update, look good and really like the detailed validation checks you have added.
Just thought it might be good to add a link to the panel you are adding support for

milkpirate:
BTW have your thought about using `enum`s for stuff like this:

- https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L57-L294
- https://github.com/weaveworks/grafanalib/blob/master/grafanalib/formatunits.py#L11

JamesGibo:
> BTW have your thought about using `enum`s for stuff like this:
> 
>     * https://github.com/weaveworks/grafanalib/blob/master/grafanalib/core.py#L57-L294
> 
>     * https://github.com/weaveworks/grafanalib/blob/master/grafanalib/formatunits.py#L11

Yeah I have thought about it, I think the issue was that when the library was started and it supported Python2.7 we wouldn't have been able to use them and now I wouldn't want to make a breaking change. If they could be added and used in parallel until a major release then I don't see why not.

JamesGibo:
Thanks again for your great work on this panel, is this PR ready to be merged? LGTM

milkpirate:
- `enum`s: Ok, I see. Well then till 1.x or 0.6.x
- :D yes, sorry for the many pushs. I just coded it and then when using it productivly I realized some missing pieces. It's mergeable now. 
- I think the next thing I'll do will be: https://grafana.com/grafana/plugins/pierosavi-imageit-panel/ just to let your know ;)

PR #412, opened 2021-10-15
Add MinMetricAgg and PercentilesMetricAgg to elasticsearch

gmmcal:
## What does this do?
I was playing around with this package and I noticed that there is no support for min and percentiles aggregations. While porting my current code to this lib, I created both classes and decided to upstream to library, so everyone can enjoy

## Why is it a good idea?
This adds more resources and power for users of the library that uses ElasticSearch as a datasource

## Questions
I noticed that ES classes has no tests. I don't know the reason, but I kept as is and did not add any. I can add, if needed


gmmcal:
@JamesGibo sure! I had double quotes because my application uses it. will apply, fix conflicts and amend commit

PR #416, opened 2021-10-20
Update grafanalib/tests/examples/example.upload-dashboard.py example to update existing dashboard

tamaszl:
added code to update existing dashboard, and to get a list of existing dashboards to retrieve id and uid.

## What does this do?

Update existing dashboard

## Why is it a good idea?

More Examples for newbies

## Context

current code only creates new dashboards. new code will also update the dashboard and get a list of current dashboards.


JamesGibo:
Contributions added in #464

PR #425, opened 2021-11-01
Add missing overrides to timeseries visualization

Puneeth-n:
## What does this do?
This PR adds the missing attributes to timeseries visualization

## Why is it a good idea?
This allows the grafanalib users to override series in Timeseries visualization

## Context
Screenshot 2021-11-01 at 9 41 15 AM


JamesGibo (approved):
Thanks for the PR, looks good to me

Issue #153, opened 2019-02-19
Feature Request: support sql queries


vbichov:
Grafana supports mysql queries out of the box.

akshayah3:
@vbichov Were you able to execute sql queries?

vbichov:
yes - I forked the repo and added the missing code:
https://github.com/vbichov/grafanalib/blob/master/grafanalib/core.py#L393

Valiev:
+1 for the support

JamesGibo:
@vbichov do you want to submit a PR to add support for this to main repo?

gardlt:
Can I make a PR against this?

rjulius23:
Ok so nobody did that so I will open it :)

rjulius23:
@JamesGibo can you please check it out ? #firsttimemaintainerhere :)

JamesGibo:
Closed in #429 
Thanks @rjulius23

rjulius23:
@JamesGibo when do you think it can be released ? It would be neat to have soon :)

JamesGibo:
@rjulius23 new release should now be published. Thanks again for your PR

PR #429, opened 2021-11-17
Add SQL Target

rjulius23:
This change is supposed to make it possible to define SQL queries as well for targets so Data Sources like MySQL, PostgreSQL can work too.

## What does this do?
Creates a New Target class named SqlTarget. With this now it is possible to define panels (Graphs) with SQL queries in them for datasources like MySQL and PostgreSQL.

## Why is it a good idea?
SQL based datasources are really widespread and with this addition now those panels can also be defined by grafanalib.

## Context
It was already mentioned in a feature request.  Feature Request: support sql queries #153 

## Questions
No questions from my side.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -426,6 +426,7 @@ class Target(object):
             'datasource': self.datasource,
         }
 
+
 @attr.s
 class SqlTarget(Target):
     """
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -426,6 +426,18 @@ class Target(object):
             'datasource': self.datasource,
         }
 
+@attr.s
+class SqlTarget(Target):
+    """
+    Metric in sql foramt
+    """
+    rawSql = attr.ib(default="")
+    def to_json_data(self):
+        super_json = super(SqlTarget, self).to_json_data()
+        super_json['rawSql'] = self.rawSql
+        super_json['rawQuery'] = True
+        return super_json
+
 
 @attr.s
 class Tooltip(object):
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -430,10 +430,16 @@ class Target(object):
 @attr.s
 class SqlTarget(Target):
     """
-    Metric in sql foramt
+    Metric target to support SQL queries
     """
     rawSql = attr.ib(default="")
+
     def to_json_data(self):
+        """Override the Target to_json_data to add additional fields.
+           rawSql: this will contain the actual SQL queries
+           rawQuery: this needs to be set to True as in case of False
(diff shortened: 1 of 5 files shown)


grafanalib/core.py
```diff
@@ -426,6 +426,18 @@ def to_json_data(self):
             'datasource': self.datasource,
         }
```
rjulius23, line 428:
```suggestion


```

rjulius23: Fix pep8

JamesGibo (changes requested):
PR looks look other than the minor linting issues.

Please could you add a test to check that the new SqlTarget class serialises and add a docstring for each of the attributes and how to use it?

I have not used SQL with grafana, but would `rawQuery` every need to be set to False? If yes, please could you make it an attribute?

rjulius23:
> PR looks look other than the minor linting issues.
> 
> Please could you add a test to check that the new SqlTarget class serialises and add a docstring for each of the attributes and how to use it?
> 
> I have not used SQL with grafana, but would `rawQuery` every need to be set to False? If yes, please could you make it an attribute?

rawQuery has to be True is you used rawSql, the reason for that is that you can also use the graphical interface to compose an SQL, but obviously it is easier in code to provide the SQL command :)

I will add test and docstrings.

rjulius23:
> PR looks look other than the minor linting issues.
> 
> Please could you add a test to check that the new SqlTarget class serialises and add a docstring for each of the attributes and how to use it?
> 
> I have not used SQL with grafana, but would `rawQuery` every need to be set to False? If yes, please could you make it an attribute?

Updated with unit test and docstrings.

rjulius23:
My bad, fixing the test.

JamesGibo:
> My bad, fixing the test.

No worries, you can test locally using tox by running `make all`

rjulius23:
> > My bad, fixing the test.
> 
> No worries, you can test locally using tox by running `make all`

Test are fixed, I also added an example that I have tested on my Grafana to be sure.

grafanalib/core.py
```diff
@@ -427,6 +427,27 @@ def to_json_data(self):
...
 
 
+@attr.s
+class SqlTarget(Target):
+    """
+    Metric target to support SQL queries
+    """
+
+    rawSql = attr.ib(default="")
+    rawQuery = attr.ib(default=True)
```
rjulius23, line 437: @JamesGibo I added the rawQuery, but in the docstring I highlighted that it should be True , now the by default is True.

JamesGibo (approved):
LGTM, thanks for the PR and dashboard example

PR #434, opened 2021-11-26
Extend SeriesOverride

milkpirate:
## What does this do?
Extends the ``SeriesOverride`` options.

## Why is it a good idea?
Especially add ``fillBelowTo`` and ``fill`` to ``SeriesOverride``.


JamesGibo (approved):
LGTM, thanks for extending this class

PR #436, opened 2021-11-30
Fix text panel

milkpirate:
## What does this do?
Fixes the TextPanel class.

## Why is it a good idea?
Makes it work again with Grafana 8.

## Questions?
Could you check if that works with other Grafana versions? The original did not wirk with 8, but I dont know if the fixed one works with <8.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,11 +11,15 @@ Changelog
 * Added overrides to panels
 * Extend ``SeriesOverride`` options
 
+Changes
+-------
+
+* Fix Text panel (and add tests)
 
 0.6.0 (2021-10-26)
 ===================
 
-* Added Discrete panel
+* Added Discrete panel (https://grafana.com/grafana/plugins/natel-discrete-panel/)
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -11,11 +11,15 @@ Changelog
 * Added overrides to panels
 * Extend ``SeriesOverride`` options
 
+Changes
+-------
+
+* Fix Text panel (and add tests)
```
JamesGibo, line 17: Please could you add a warning that the fix has only been tested on > Grafana v8

JamesGibo (changes requested):
Thanks for the PR and LGTM

Regarding testing on < Grafana v8, currently I only run Grafana v8 so would a bit of pain to test. I personally have no issue with merging this change as we have released v0.6.0, which has breaking changes for previous version of Grafana. I would just add a warning in the changelog and users can pin to a older version if they have issues.

What do you think @dholbach?

dholbach:
Yes, that sounds good to me.

JamesGibo (approved):

PR #437, opened 2021-12-01
Add threshold type option for thresholds

STandon-3:
## What does this do?
Adds a `thresholdType` to the `Panel` class.

## Why is it a good idea?
Allows users to choose between an absolute or percentage type for thresholds (previously only absolute types were possible), which is useful for visualisations.

## Context
Have moved the `thresholds` argument to the `Panel` class to make this implementation neater.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Changelog
 * Added missing attributes to the Logs panel
 * Added Cloudwatch Logs Insights Target
 * Added overrides to panels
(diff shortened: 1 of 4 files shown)


JamesGibo:
I helped to co-author this PR, so @dholbach please could you just give it a quick look over to check I have not missed anything?
The default functionality for panels should not have changed

dholbach (approved):
This looks great to me.

JamesGibo (approved):

PR #442, opened 2021-12-07
Datasource parameter for CloudWatch targets

romich:
## What does this do?
Adds the ability to set datasource for CloudWatch targets, like in regular `Target`.

## Why is it a good idea?
So that CloudWatch metrics could be used in panels with mixed datasources.


JamesGibo (approved):
LGTM, thanks for the contribution!

PR #448, opened 2022-01-07
Add support for fields value in Stat panel

gkuga:
## What does this do?

Added fileds to Stat panel because it was not possible to set it.

## Context

Where values were not displayed in the Stat panel, fields needed to be set.


JamesGibo (approved):
Thanks for the PR, looks good to me

PR #449, opened 2022-01-07
alertName parameter for AlertList panel

romich:
## What does this do?
Adds the ability to filter alerts in AlertList panel.
Screenshot 2022-01-07 at 15 17 44

## Why is it a good idea?
Because filtering doesn't work right now.

## Context
Not sure if it is because of recent Grafana version, or using new grafana alerting system, but `AlertList` panels created manually have quite different model from the ones generated by grafanalib, and not all parameters work for it.

## Questions
I don't know what is the correct course of action would be here, but for now I've just fixed one parameter that I need (hopefully in a non-breaking way).


JamesGibo (approved):
LGTM

PR #453, opened 2022-01-30
Add `to_json_data` for `Repeat` class

achew22:
## What does this do?

Previously it wasn't possible to make a Repeat class because it couldn't be turned into JSON

## Why is it a good idea?

This makes it possible

## Context

I think this is reasonably straight forward, but I'm happy to provide more context if you have questions.

## Questions

No questions, thank you!


JamesGibo (approved):
Thanks for the PR not sure how that one got missed.
We should probably add a test for it but don't expect you to do that unless you wanted to?

PR #454, opened 2022-02-06
Extend SeriesOverride (dashes, Z-index)

milkpirate:
## What does this do?
Extends the SeriesOverride options by dashes and Z-index.

## Why is it a good idea?
More functionality.


milkpirate:
Dont know, what to do about: `Fail if there were link errors` :confused:

PR #457, opened 2022-02-16
Add thresholdsStyleMode parameter for TimeSeries panel

gkuga:
## What does this do?

Adds the ability to set thresholdsStyleMode in TimeSeries panel.
スクリーンショット 2022-02-16 15 49 25

## Why is it a good idea?

Because the threshold drawing style could not be set.

## Context

In the TimeSeries panel, the threshold drawing style could not be set.


gkuga:
How about removing the change log from the URL linting target since the contents easy to out of date?

JamesGibo (approved):
Thanks for your PR, look good!
I will try to make a release to include it by the end of the week.

PR #471, opened 2022-03-10
Add support to set legend values on TimeSeries

jlusiardi:
## What does this do?
It adds a new parameter `legendCalcs` to the TimeSeries panel to define which calculations are shown in the legend. 

## Why is it a good idea?
We were missing this feature duration the creation of out dashboards and think other user would also like to use it.

## Context
Only introduces the new parameter, no big thing ;)

## Questions
None

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* Added ``legendCalcs`` parameter to TimeSeries Panel
+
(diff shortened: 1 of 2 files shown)


jlusiardi:
Just out of curiosity, is there anything more i can/must do here?

jlusiardi:
No Problem, i just wasn't sure if I missed something I had to do.

PR #477, opened 2022-03-24
added hide field for ElasticSearchTarget and added ExpressionTarget s…

pdarvasi-epam:
## What does this do?
The hide option lets us create panels where we can hide some of the queries that we have no intention to show. In combination with the added ExpressionTarget we can display only the calculated values.

## Why is it a good idea?
It provides the ability to do calculations with multiple results, e.g. calculate service health rate based on all responses and 2xx/3xx responses.

## Context
The added json is supported by Grafana, it can be created by hand.

## Questions
No questions. Tested with temporary workaround classes in our project and it works. Note: refID can be used as an alias for the ExpressionTarget.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+0.6.2 (2022-02-24)
+==================
+
 * Added percentage type for thresholds
 * Added ``datasource`` parameter to CloudWatch targets
 * Added support for auto panels ids to AlertList panel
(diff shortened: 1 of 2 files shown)


JamesGibo (approved):
Thanks for the PR, look really good.
Would you be able to rebase against main and if possible add a test for the ExpressionTarget Class you have added in new file called grafanalib/tests/test_elasticsearch.py?

pdarvasi-epam:
Sure, let me create the changes and update the PR.

pdarvasi-epam:
https://github.com/weaveworks/grafanalib/pull/483 opened, as this one got auto closed.

PR #480, opened 2022-03-30
Add Azure Monitor Targets

cam-barts:
## What does this do?
Added support  for the [Azure Monitor Datasource](https://grafana.com/docs/grafana/latest/datasources/azuremonitor/#azure-monitor-data-source) as a target.

## Why is it a good idea?

Azure monitor is heavily utilized in my corporate environment at least as a data source, and I suspect in many others as well. Because many of our services are deployed in Azure, we'd love the ability to create dashboards programatically via grafanalib. This contribution is a way for us to give back to the tool we will be heavily utilizing in the near future. 

## Context

The azure monitor data source exposes 3 types of data to query: Azure Monitor Metrics, Azure Logs, and Azure Resource Graph. They are all outlined on the linked document. I wrote a target and individual tests for each of those types.

## Questions

I am pretty certain I'm adhering to the style guides, but my editor runs black on save so there might have been changes that that made that don't align exactly. I didn't spot anything on a cursory look, but if there is anything egregious I am happy to fix.


JamesGibo (approved):
Thanks for the PR, I think this will be really useful to other people.
Look good to me although I do not use Azure so cannot test.
Please could you rebase against main?

PR #491, opened 2022-05-13
add 'datasource' parameter for InfluxdbTarget

sbng:
Under each InfluxdbTarget, a different data source can be specified. This allow the merging of trends from different influxb into the same Panel. This is done by adding the 'datasource' parameter to the InfluxdbTarget object

## What does this do?
Add a datasource for different InfluxdbTarget. Without this option, the InfluxdbTarget could only have one default datasource specify outside Target.

## Why is it a good idea?
Different datasource for each Influxdbtarget allows the merging of different trend onto the same Panel.

## Context
The current setup using dataSource specify outside the target stencils. Adding a data source within the target allows a combination of different data sources on the same panel. This is carried out via adding "datasource" parameter within the influxdbTarget object similar to adding the parameters like query and measurement. This standalone datasource is supported by grafana.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ x.x.x (TBC)
 ===========
 
 * Added ...
+* Added datasource parameter to Influxdb targets
 
 
 0.6.3 (2022-03-30)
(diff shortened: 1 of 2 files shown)


JamesGibo:
Thanks for the the PR, sure others will find that feature really useful.
The PR looks good to me. 
Do not worry about the failed CI check, the Grafana documentations links seems to have changed, I will fix that in a separate PR. 

Please could you add a note about what functionality you have added to the Changelog file.

Thanks

sbng:
Updated the Changelog file as per request. Thank you.

PR #494, opened 2022-05-20
Add missing format units

mharbison72:
## What does this do?
This starts by adding the new Boolean unit formats, but also adds other formats that I noticed were missing.  There are likely still some missing, like in the currency section.

## Why is it a good idea?
Fewer missing options, which are available when manually creating a dashboard.

## Questions
It would sure be nice to auto-generate this file by grabbing the upstream file and processing it, but I don't have any good ideas for generating nice names (or keeping the current constant names).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ x.x.x (TBC)
 
 * Added ...
 * Added datasource parameter to Influxdb targets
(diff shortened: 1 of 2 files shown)


JamesGibo (approved):
Thanks for adding the missing units that is really useful.
It would be nice to autogenerate the list, but probably not worth the effort as they don't change that often and can be manually overridden to any unit not supported.

PR #496, opened 2022-05-23
Small cleanup

mharbison72:
## What does this do?
These are small cleanups for lint type things flagged by PyCharm.  I didn't bother with updating the changelog because AFAICT, it won't have user visible changes, but still is more correct.

## Why is it a good idea?
Fewer lint type things get flagged in the IDE


JamesGibo (approved):
Well spotted, thanks for the contribution

PR #520, opened 2022-08-10
fix: target validation

OscarVanL:
## What does this do?
Target validation requires that refId is not blank. This is because an AlertCondition needs to apply to a target refId.

However, the current validation does not work if the refId is omitted. This is because it is defaulted to "" if it's missing, but the previous check was just for `hasattr`.

## Why is it a good idea?

It prevents people making invalid alerts that do not apply to any target. This could be a source of bugs.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -391,10 +391,10 @@ class Repeat(object):
 
 def is_valid_target(instance, attribute, value):
     """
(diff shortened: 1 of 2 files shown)


JamesGibo:
Thanks for the PR, I think that check makes sense will merge now.

PR #522, opened 2022-08-11
fix: add missing thresholds param to Stat json

OscarVanL:
## What does this do?

This adds the thresholds param (which already existed in the Stat panel's params) into the generated JSON

## Why is it a good idea?

Thresholds are not applied to the generated JSON at the moment.


OscarVanL:
Nevermind, this did work, I just misunderstood the code!

PR #528, opened 2022-09-09
Add support for Ae3e Plotly panel

trewallier:
## What does this do?
Add support for [Ae3e Plotly panel](https://github.com/ae3e/ae3e-plotly-panel)

## Why is it a good idea?
Add support for a widely used panel type [Downloads](https://grafana.com/grafana/plugins/ae3e-plotly-panel)

## Context
The panel use two json files (dict in python) to store the configuration, and strings to store the display scripts. This makes it highly flexible, as it only depends on the high level structure of plotly. So the panel integration does the same, and does not do extra validation of the setup jsons.

##  Extra note
Since config and layout is added directly into the json schema, they will not match in the json output. Mind the subtle differences between python dict and json format (eg. `False` <-> `false`)


JamesGibo (approved):
Thanks for the PR, looks like a really useful panel that I have not used before.
Thanks also for adding some tests for the panel.
Will merge now and do a release towards the end of the week.

PR #543, opened 2022-11-09
Add support for `statistic` field in `CloudwatchMetricsTarget`

Butterneck:
## What does this do?
Add support for `statistic` field in `CloudwatchMetricsTarget`

## Why is it a good idea?
The existing `statistics` field is being ignored by grafana 9.x, it must be named `statistic` (not `statistics`) and must be a string (not a list)

### DIFF
--- a/grafanalib/cloudwatch.py
+++ b/grafanalib/cloudwatch.py
@@ -26,7 +26,7 @@ class CloudwatchMetricsTarget(object):
     :param period: Cloudwatch data period
     :param refId: target reference id
     :param region: Cloudwatch region
-    :param statistics: Cloudwatch mathematic statistics
+    :param statistics: Cloudwatch mathematic statistics (to be deprecated, prefer `statistic` instead)
     :param statistic: Cloudwatch mathematic statistic
     :param hide: controls if given metric is displayed on visualization
     :param datasource: Grafana datasource name
--- a/grafanalib/cloudwatch.py
+++ b/grafanalib/cloudwatch.py
@@ -26,7 +26,8 @@ class CloudwatchMetricsTarget(object):
     :param period: Cloudwatch data period
     :param refId: target reference id
     :param region: Cloudwatch region
-    :param statistics: Cloudwatch mathematic statistic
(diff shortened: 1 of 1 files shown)


grafanalib/cloudwatch.py
```diff
@@ -26,7 +26,8 @@ class CloudwatchMetricsTarget(object):
     :param period: Cloudwatch data period
     :param refId: target reference id
     :param region: Cloudwatch region
-    :param statistics: Cloudwatch mathematic statistic
+    :param statistics: Cloudwatch mathematic statistics
+    :param statistic: Cloudwatch mathematic statistic
```
JamesGibo, line 30: Please could you add a note about why there are 2 fields ( < v9 and > v9 for example) and the difference in formatting (list vs string)?

JamesGibo (changes requested):
Thanks for this PR, I do not use cloudwatch so will have to trust that this is required, but as this is not a breaking change I am happy to add it.
Please could you add a little description of what the difference is between the 2 fields to the doc strings?

grafanalib/cloudwatch.py
```diff
@@ -26,7 +26,8 @@ class CloudwatchMetricsTarget(object):
     :param period: Cloudwatch data period
     :param refId: target reference id
     :param region: Cloudwatch region
-    :param statistics: Cloudwatch mathematic statistic
+    :param statistics: Cloudwatch mathematic statistics
```
tiansu-yu-idealo, line 29: Unsolicited advise: Should there be a message e.g. `(to be deprecated, prefer `statistic` instead)`

JamesGibo, line 29:
```suggestion
    :param statistics: Cloudwatch mathematic statistics (to be deprecated, prefer `statistic` instead)
```

JamesGibo (changes requested):

grafanalib/cloudwatch.py
JamesGibo: Added a note to show `statistics` deprecated

JamesGibo (approved):

PR #544, opened 2022-11-11
fix: AlertRulev9 JSON schema is not correct

OscarVanL:
There are a couple of mistakes in the Grafana 9 Alerts-As-Code `AlertRulev9` JSON output.

Firstly, the absence of `grafana_alert` means that uploading the output JSON fails with a validation error: `{"message":"unexpected backend type (grafana) for payload type (lotex)","traceID":""}`, which specifically is [this error in Grafana](https://github.com/grafana/grafana/blob/718620c197614f94dd15231790fc6530c3cc8b8a/pkg/services/ngalert/api/errors.go#L20-L25).

Secondly, the fields `noDataState` and `execErrState` were renamed to `no_data_state` and `exec_err_state` in v9, meaning these settings were not applied correctly.

With these changes my alerts upload successfully and appear to be correct, but if I spot further mistakes I'll make follow-up PRs.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1625,15 +1625,17 @@ class AlertRulev9(object):
                 data += [trigger.to_json_data()]
 
         return {
-            "title": self.title,
             "uid": self.uid,
-            "condition": self.condition,
             "for": self.evaluateFor,
             "labels": self.labels,
             "annotations": self.annotations,
-            "data": data,
-            "noDataState": self.noDataAlertState,
-            "execErrState": self.errorAlertState
+            "grafana_alert": {
+                "title": self.title,
+                "condition": self.condition,
+                "data": data,
+                "no_data_state": self.noDataAlertState,
+                "exec_err_state": self.errorAlertState,
+            },
         }
 
 
--- a/grafanalib/tests/test_core.py
+++ b/grafanalib/tests/test_core.py
@@ -901,11 +901,11 @@ def test_alertrulev9():
     )
 
     data = rule.to_json_data()
-    assert data['title'] == title
     assert data['annotations'] == annotations
     assert data['labels'] == labels
     assert data['for'] == "3m"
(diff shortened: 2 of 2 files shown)


JamesGibo:
Thanks for the PR, I will give this a test with my alert rules as I have not seen this issue, but then I am not uploading via the API.

JamesGibo:
> Interesting, how are you uploading the alerts? By provisioning?

Yeah I use the file based provisioning.

OscarVanL:
This looks to be the format that Grafana uses internally, if you open Inspect Element and go to the Alerting tab of Grafana, then search for the request to `/api/ruler/grafana/api/v1/rules?subtype=cortex`, and look at the response body, you can see the JSON structure. For me (Grafana 9.2.3), it matches the structure that this PR uses.

Out of curiosity, if you find one of the alerts you uploaded using file-based provisioning inside this response, what structure does it use?

hugohaa:
Just a FYI I also tested this PR in our project and it's working perfectly. Was having a hard time figuring out what was happened. Thanks!

OscarVanL:
Thanks @hugohaa! Is there any possibility I could get a second look at this PR @JamesGibo?

ran-ka:
We are also using the API to upload our alerts, and had to merge this PR for it to work. Thanks @OscarVanL

JDuskey:
This looks good, and can confirm it fixes errors when pushing alerts to the unified alerting API.

JamesGibo (approved):
Looks good to me, I have not tested the updated JSON format, but will test it before publishing a release

soheil-ta:
@JamesGibo Is there any plan to release the latest version including this fix?
The latest version is 0.7.0 from Nov 2022 which does not include the fix.

urosht:
+1 here, any idea on the next release for this, @JamesGibo?

PR #554, opened 2022-11-26
Expose Now delay option in TimePicker

OscarVanL:
## What does this do?

The Now delay can be applied to a dashboard (via a TimePicker), which will delay the time at which `now` applies when looking at data on a dashboard.
image

image

## Why is it a good idea?

This was missing from Grafanalib.


JamesGibo (approved):
LGTM, thanks for the PR

PR #555, opened 2022-11-28
Update SqlTarget to support parsing queries from files

kamka427:
## What does this do?
Add the ability to parse SQL queries from files.

## Why is it a good idea?
Parsing files instead of using query strings reduces code length and improves code readability.

## Context
Complex panels and dashboards use multiple long queries. Parsing them from files shortens the code length and leads to a more readable and maintainable codebase. The implementation supports parametrizing the queries with dictionary as an attribute to help writing queries with variables.


JamesGibo (approved):
Thanks for the PR, I can see this being a really useful feature.
May even extend this support to other datasource types.

PR #560, opened 2022-12-27
fix: AlertRulev9 does not support CloudwatchMetricsTarget

ran-ka:
## What does this do?
Some small fixes for Grafana 9 alerts support:
- Grafana9 AlertExpression accepts only Targets while CloudwatchMetricsTarget for some reason is not a Target
- AlertCondition can include multiple conditions, each with its own target

## Why is it a good idea?
- Maintain support for CloudwatchMetricsTarget in alerts
- Support the full complexity of the implementation of the new alert

### DIFF
--- a/grafanalib/cloudwatch.py
+++ b/grafanalib/cloudwatch.py
@@ -3,10 +3,11 @@
 import attr
 
 from attr.validators import instance_of
+from grafanalib.core import Target
 
 
 @attr.s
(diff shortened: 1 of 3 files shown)


JamesGibo (approved):
PR looks good, thanks for contributing.
I was worried that by inheriting from `Target` class that there would be a large number of unnecessary fields included in the output JSON, but this is not the case as you do not call `super`

PR #568, opened 2023-01-27
Fix rendering AlertRulev9 json for possible use in Alerting provisioning API

ilia-ilin-jiji:
## What does this do?

This PR fix generates the correct json for HTTP API requests.

1. Add link alert with dashboard and panel. In AlertRulev9 class exist params `panel_id` and `dashboard_uid` but didn't use in json generating.
2. Fix defaults for `labels` and `annotations` params because are work incorrectly. If I change param in one alert is changed in all
```
>>> import attr
>>> @attr.s
... class X:
...     d = attr.ib(default={})
...
>>> a = X()
>>> b = X()
>>> a.d['test']='test'
>>> b
X(d={'test': 'test'})
>>>
```

3. Add DataSource class for correct generating targets in alert_rule, because API need this struct

```

## Why is it a good idea?
<!-- how does it help grafanalib users / maintainers? -->

## Context
<!-- any background that might help the reviewer understand what's going on -->

## Questions
<!-- things you're uncertain about that you want the reviewer to focus on -->

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -842,18 +842,9 @@ class DataSourceInput(object):
 
 
 @attr.s
-class DataSourceForTargets(object):
+class DataSource(object):
     uid = attr.ib(validator=instance_of(str))
-    type = attr.ib(
-        validator=in_([
-            PLUGIN_ID_CLOUDWATCH,
-            PLUGIN_ID_GRAPHITE,
-            PLUGIN_ID_INFLUXDB,
-            PLUGIN_ID_OPENTSDB,
-            PLUGIN_ID_PROMETHEUS,
-            PLUGIN_ID_ELASTICSEARCH,
-        ])
-    )
+    type = attr.ib(validator=instance_of(str))
 
     def to_json_data(self):
         return {
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1600,8 +1600,8 @@ class AlertRulev9(object):
 
     title = attr.ib()
     triggers = attr.ib(default=[], validator=is_valid_triggersv9)
-    annotations = attr.ib(default={}, validator=instance_of(dict))
-    labels = attr.ib(default={}, validator=instance_of(dict))
+    annotations = attr.ib(factory=dict, validator=instance_of(dict))
+    labels = attr.ib(factory=dict, validator=instance_of(dict))
     folderUid = attr.ib(default=None, validator=attr.validators.optional(instance_of(str)))
     ruleGroup = attr.ib(default=None, validator=attr.validators.optional(instance_of(str)))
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by JDuskey** [comment]
#544 Has the correct implementation of this fix. @KacperLegowski is right that alerts need a nested grafana_alert object.

File: grafanalib/core.py [hunk 1]
```diff
@@ -1633,7 +1653,9 @@ def to_json_data(self):
             "annotations": self.annotations,
             "data": data,
             "noDataState": self.noDataAlertState,
-            "execErrState": self.errorAlertState
+            "execErrState": self.errorAlertState,
```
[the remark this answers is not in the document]
JamesGibo: Please could you remove this part from the PR

grafanalib/core.py
```diff
@@ -1574,13 +1586,15 @@ class AlertRulev9(object):
...
     """
 
     title = attr.ib()
     triggers = attr.ib(default=[], validator=is_valid_triggersv9)
-    annotations = attr.ib(default={}, validator=instance_of(dict))
-    labels = attr.ib(default={}, validator=instance_of(dict))
+    annotations = attr.ib(factory=dict, validator=instance_of(dict))
+    labels = attr.ib(factory=dict, validator=instance_of(dict))
+    folderUid = attr.ib(default=None, validator=attr.validators.optional(instance_of(str)))
+    ruleGroup = attr.ib(default=None, validator=attr.validators.optional(instance_of(str)))
```
JamesGibo, line 1597: Add document string for these

JamesGibo (changes requested):
Thanks for the PR, there looks like there are some really useful improvements in the PR.
Would you be bale to-do the following:

- [ ] rebase on master
- [ ] Remove the incorrect 
- [ ] Add documentation strings for the additional fields you have added

grafanalib/core.py
```diff
@@ -1574,13 +1586,15 @@ class AlertRulev9(object):
     :param timeRangeTo: Time range interpolation data finish at
     :param uid: Alert UID should be unique
     :param dashboard_uid: Dashboard UID that should be use for linking on alert message
-    :param panel_id: Panel ID that should should be use for linking on alert message
+    :param panel_id: Panel ID that should be use for linking on alert message
```
Gnappuraz, line 1589:
```suggestion
    :param panel_id: Panel ID that should be used for linking on alert message
    :param annotations: An array of additional annotations, can be used for alert message
    :param labels: An array of labels, can be used with Notification Policies
    :param folderUid: Folder ID that should be use for linking on alert message
    :param ruleGroup: A group for alerts that are evaluated together
```

Gnappuraz (commented):
I would be happy to help out on this. I'm currently trying to use the lib to implement alert provisioning via the new API but I'm struggling because of this and other issues.

Gnappuraz (commented):
I would be happy to help out on this. I'm currently trying to use the lib to implement alert provisioning via the new API but I'm struggling because of this and other issues.

PR #583, opened 2023-05-07
Add support for Elasticsearch Rate Metric Aggregator

jimbolimbo13:
## What does this do?

This adds support for the "Rate Aggregation" as a metric in an Elasticsearch query.

image

## Why is it a good idea?

This is supported in Grafana, and my team needed it for something. I figure others might too.

### DIFF
--- a/grafanalib/elasticsearch.py
+++ b/grafanalib/elasticsearch.py
@@ -2,7 +2,7 @@
 
 import attr
 import itertools
-from attr.validators import instance_of
+from attr.validators import in_, instance_of
 from grafanalib.core import AlertCondition
 
 DATE_HISTOGRAM_DEFAULT_FIELD = 'time_iso8601'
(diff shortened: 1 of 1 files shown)


grafanalib/elasticsearch.py
```diff
@@ -498,3 +498,48 @@ def to_json_data(self):
             'inlineScript': self.inline,
             'settings': self.settings,
         }
+
```
JamesGibo, line 501:
```suggestion


```

JamesGibo: Fix linting error

JamesGibo (changes requested):
Thanks for the PR, it looks good, just a minor import missing.

Add the following to fix the test error
```
from attr.validators import in_
```

Please could you also add a note to the release notes?

I have also added an issue to test the Elasticsearch file as realised it is missing a test file. 
https://github.com/weaveworks/grafanalib/issues/589

JamesGibo (approved):
LGTM

jimbolimbo13:
haha, you were too quick. I just pushed the changelog change, but it was missed

Issue #589, opened 2023-05-18
Add basic tests for Elasticsearch 


JamesGibo:
Add tests to check Elasticsearch classes, to check they serialise to JSON
https://github.com/weaveworks/grafanalib/blob/main/grafanalib/elasticsearch.py

Tests can be templated from any of the other test files in this folder:
https://github.com/weaveworks/grafanalib/tree/main/grafanalib/tests

jimbolimbo13:
I can kick this off by making a test for the class I added

PR #594, opened 2023-05-18
Add unit test for RateMetricAgg in elasticsearch

jimbolimbo13:
## What does this do?

Adds a file for unit tests on the elasticsearch module
Adds a unit test for the RateMetricAgg class added in #583 

## Why is it a good idea?

Unit tests are a good thing. I found a missing parameter implementation when making this (the fix for that is also part of this PR)

## Context

https://github.com/weaveworks/grafanalib/issues/589


JamesGibo (approved):
Thanks for adding the tests

PR #602, opened 2023-06-11
Add color support in Stat(panel)

joewangprc:
## What does this do?
Currently Stat(panel) doesn't support color option, so it still always be grey color when using Stat(panel)

## Why is it a good idea?
Have a choice of color improve the readable of panel

## Context
Color option is simpler than thresholds, in the case of no need to define thresholds

## Questions
not sure about the combination, in my case, I always use together with colorMode="background" and graphMode="none"


JamesGibo (approved):
LGTM, thanks for you contribution

PR #610, opened 2023-07-25
Update python compatibility in README

emirkmo:
## What does this do?

Update max python version compatibility in readme to include python 3.11. 

## Why is it a good idea?

3.11 is part of tests so README should not mislead about only supporting up to python 3.10
## Context

See passing tests https://github.com/weaveworks/grafanalib/actions/runs/5575606128/job/15097310698. It is also published to pypi.

## Questions

I don't think such a minor docs fix needs a CHANGELOG entry, correct?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,7 @@ Support
 This library is in its very early stages. We'll probably make changes that
 break backwards compatibility, although we'll try hard not to.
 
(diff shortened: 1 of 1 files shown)


JamesGibo (approved):
LGTM, thanks for the update

PR #612, opened 2023-07-31
Added sortBy parameter to Table panel

Git-Lior:
## What does this do?
This change allows sorting table rows by their fields

## Why is it a good idea?
This is a useful feature which is supported in grafana v8:
https://grafana.com/docs/grafana/v8.0/panels/visualizations/table/#sort-column

## Context
`sortBy` in grafana's Table panel options:
https://github.com/grafana/grafana/blob/v8.0.x/public/app/plugins/panel/table/models.gen.ts#L17

The name for `TableSortByField` was inspired by the matching type in grafana:
https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Table/types.ts#L24

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -3186,6 +3186,18 @@ class Column(object):
 
 
 @attr.s
+class TableSortByField(object):
+    displayName = attr.ib(default="")
+    desc = attr.ib(default=False)
+
+    def to_json_data(self):
+        return {
+            'displayName': self.displayName,
+            'desc': self.desc,
+        }
+
+
+@attr.s
 class Table(Panel):
     """Generates Table panel json structure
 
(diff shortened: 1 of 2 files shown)


grafanalib/core.py
```diff
@@ -3210,6 +3223,7 @@ class Table(Panel):
     showHeader = attr.ib(default=True, validator=instance_of(bool))
     span = attr.ib(default=6),
     unit = attr.ib(default='', validator=instance_of(str))
+    sortBy = attr.ib(default=attr.Factory(list))
```
JamesGibo, line 3226:
```suggestion
    sortBy = attr.ib(default=attr.Factory(list), validator=attr.validators.deep_iterable(
        member_validator=instance_of(TableSortByField),
        iterable_validator=instance_of(list)
    ))
```

JamesGibo (changes requested):
PR looks good, thanks for your contribution.
I think it would be good to add a validator to the `sortBy` variable and have added a suggestion for you to test

JamesGibo (approved):
LGTM

Issue #540, opened 2022-11-04
'total' literal not correct for Gauge Calculation


ssmcgrat:
https://github.com/weaveworks/grafanalib/blob/ff156bb13dc3fa6b5df2abc8b5e62bde4ab8935e/grafanalib/core.py#L278

Howdy, I believe `GAUGE_CALC_TOTAL = 'total'` needs to be changed to `GAUGE_CALC_TOTAL = 'sum'`. The literal 'total' fails to define the Calculation value for a bar gauge, 'sum' literal has worked for me (using grafanalib 0.7.0)

davaffy:
I can confirm this:

![image](https://github.com/weaveworks/grafanalib/assets/58271442/fb83ecaf-11a6-4145-8ed2-4d176bbcca72)

This can also be seen here where total is set to sum.

https://github.com/grafana/grafana/blob/71b841fa9bfbc02f40f23a98d0490577603aba02/packages/grafana-data/src/transformations/fieldReducer.ts#L172

I've used this case to create my first pull request (because it was an easy one)

PR #614, opened 2023-08-01
GAUGE_CALC_TOTAL should be 'sum' and not 'total'

davaffy:
## What does this do?
Correct wrong literal value for total values

## Why is it a good idea?
It's better that the users use the literals defined by grafanalib instead of their own so that future changes are autmatically adopted.

## Context
#540 is corrected with this.

## Questions
None


JamesGibo (approved):
Good spot, thanks for the PR

PR #617, opened 2023-08-17
Added axisSoftMin and axisSoftMax options to TimeSeries

DP3t3r:
## What does this do?
added axisSoftMin and axisSoftMax options to TimeSeries

## Why is it a good idea?
These options were missing

## Context
added axisSoftMin and axisSoftMax options to TimeSeries

## Questions
n/a


JamesGibo (approved):
Thanks for adding the missing parameters.

PR #622, opened 2023-08-30
Main improved dashboard links

padraic-padraic:
## What does this do?
Extends the implementation of DashboardLink to support creating both direct links and lists of dashboards.

## Why is it a good idea?
This enables grafanalib users to fully employ the DashboardLink feature. The previous JSON model was no longer accurate for recent grafana versions, meaning dashboard links were not properly generated. Here we add all available fields, allowing users to create dashboard links correctly from grafanalib.

## Context
Dashboard Links are a way to create links along the top of a dashboard, linking either to other sets of dashboards from the same organization or arbitrary external resources. See the documentation here: https://grafana.com/docs/grafana/latest/dashboards/build-dashboards/manage-dashboard-links/#manage-dashboard-links

## Questions
I have removed a parameter from the old DashboardLink class, `dashboard`; this could consistute a breaking change.

This parameter was not actually consumed by Grafana, so was in effect only acting as an internal alias to the `title` field. The `title` would only be shown if the `DashboardLink.type` attribute is set to `link`, something actually not possible before this PR. so I don't consider it to be a significant change. But, if desired, we can retain that alias-like behaviour to maintain API compatability.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -12,7 +12,7 @@ import math
 import string
 import warnings
 from numbers import Number
-from typing import Literal, Optional
+from typing import Literal
 
 import attr
 from attr.validators import in_, instance_of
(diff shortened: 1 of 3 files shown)


padraic-padraic:
Thanks for approving the workflows, will work to correct the tests shortly!

padraic-padraic:
Looks like the issue is with py37/38 support, will push some fixes shortly.

padraic-padraic:
Running locally against Python 3.8 passes now, screenshot is provided 🙂 
Screenshot 2023-09-08 at 10 50 57

grafanalib/core.py
```diff
@@ -74,7 +75,7 @@ def to_json_data(self):
 FLOT = 'flot'
 
 ABSOLUTE_TYPE = 'absolute'
-DASHBOARD_TYPE = 'dashboard'
+DASHBOARD_TYPE = Literal['dashboards', 'link']
```
JamesGibo, line 78: Should this be `dashboard`, as this is what is set as the default value in the Class?

```suggestion
DASHBOARD_TYPE = Literal['dashboard', 'link']
```

JamesGibo (changes requested):
Thanks for the PR it looks really good and well documented.
I spotted one issue that might be type that I would like you to check. 

If you had time, you could add a test for this class to the tests, but as this Class did not previously have a test, I am happy to merge without this.

grafanalib/core.py
padraic-padraic: Ah, good catch; actually it's the default that should be changed, attached here is a screenshot of some actual dashboard json generated using the Grafana UI. That's my bad!

I'm happy to add some tests, will get on that asap. 
Screenshot 2023-09-11 at 15 27 53

JamesGibo (approved):
LGTM

PR #632, opened 2023-09-29
Main add graph unit

padraic-padraic:
With thanks to my colleagues Damiano Gianotti (https://github.com/D3mianGianoz) and Ianthe Michiels.

## What does this do?
Adds the option to specify the unit for a standard `core.Graph` panel. This feature was implemented for the `core.TimeSeries` panel, but not yet for `Graph`.

## Why is it a good idea?
It brings a bit of feature parity between the two standard panel types.

## Context
For older grafana versions, Graph is the default visualisation, and is still supported (if deprecated). Having feature parity between the two panel types makes it easier to then transition to the TimeSeries panel in the future.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 ==================
 
 * Added ...
+* Add Unit option for Graph panel
 * Added Minimum option for Timeseries
 * Added Maximum option for Timeseries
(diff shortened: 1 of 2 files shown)


JamesGibo (approved):
Thanks for your contribution, the PR looks good to me!

Issue #627, opened 2023-09-08
Alert compatibility with Grafana v10


kosi96:
Hello everyone,

firstly I would like to thank all the contributors for doing such great work. 
Having dashboards and alerts in python code is so much more maintainable than doing it in the UI window.

As I am currently using Grafana Cloud, my version is v10. I am having some issues when attempting to generate and publish new alerts. I suspect that the structure of alerts has changed a little bit from v9 to v10, which may require some manual JSON adjustments to make things work (generating a proper JSON).

Does somebody know if Grafana v10 alerts will be supported? As it is for `AlertRulev8` and `AlertRulev9`.

PR #645, opened 2024-01-25
feat: add grafana v10 AlertRule support

desmaraisp:
## What does this do?

Grafana v10 has a slightly different format for alert rules, so the v9 version of the class cannot be used without overriding the to_json_data method. The PR implements the modified rendering that should work for grafana v10 alerts

## Why is it a good idea?

Supporting v10 will help keeping the library relevant in the short term while [cog](https://github.com/grafana/cog) matures. Once cog become mature enough, it might be an interesting option, but for now manual code changes are still the best way to go forward

## Context

See #627

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -195,6 +195,7 @@ EXP_TYPE_CLASSIC = 'classic_conditions'
 EXP_TYPE_REDUCE = 'reduce'
 EXP_TYPE_RESAMPLE = 'resample'
 EXP_TYPE_MATH = 'math'
+EXP_TYPE_THRESHOLD = 'threshold'
 
 # Alert Expression Reducer Function
 EXP_REDUCER_FUNC_MIN = 'min'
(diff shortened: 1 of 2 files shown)


lexi-k:
Hi, would it be possible to merge this please? We would also appreciate this change, although it hard to make the change ourselves, having it right in the "main" library would not trigger our "OCDs" as much ;-)

pawel-biegun-sq3:
Hi @amotl, would it be possible to merge this? It looks like an essential improvement to support V10 dashboards.

amotl:
Hi Pawel. I am not a member of the grafanalib project, so I can't help you here. Apologies.

garett-eiq:
@JamesGibo any chance this might get merged soon? I'd be glad to help out in some way

lordsoffallen:
Can we please merge this? Any blockers?

desmaraisp:
Fyi I would suggest trying the grafana foundation sdk instead, it's now in public preview. If it works for your usecase, you can avoid a whole bunch of issues like this one

Issue #650, opened 2024-02-14
grafanalib 0.7.1 trips with `ImportError: cannot import name 'Literal' from 'typing'`


amotl:
Hi there,

first things first: Thanks a stack for conceiving and maintaining grafanalib.

Looking at our [CI run #7525555866](https://github.com/panodata/grafana-wtf/actions/runs/7525555866/job/20482088491#step:5:32), it trips like this:

```python
ImportError while loading conftest '/home/runner/work/grafana-wtf/grafana-wtf/tests/conftest.py'.
tests/conftest.py:9: in <module>
    import grafanalib.core
.venv/lib/python3.7/site-packages/grafanalib/core.py:15: in <module>
    from typing import Literal
E   ImportError: cannot import name 'Literal' from 'typing' (/opt/hostedtoolcache/Python/3.7.17/x64/lib/python3.7/typing.py)
make: *** [Makefile:33: test] Error 4
```

We did not investigate about the root cause yet, just wanted to report here for a start.

With kind regards,
Andreas.

amotl:
Most probably, Python 3.7 became unsupported without bumping to 0.8.0. That would be the easiest and most straight-forward explanation. Is it true?

davaffy:
The reason is because of commit #622 which added the Literal import from typing.
That is not available in 3.7 thus given error.

The changelog says:

* Correct type annotations for py37, py38

However, it breaks py37 functionality. Is it not possible to upgrade to 3.8, which has it (since 3.7 is EOL) ?

It's my opinion that either the Literal has to be changed in order to make it back compliant with 3.7 ( by using typing_extensions) or 3.7 should be removed as supported version and you will need to upgrade.

Ideally it should be the second one in my opinion

amotl:
Dear @davaffy,

thanks for your quick reply.

> [...] or 3.7 should be removed as supported version and you will need to upgrade.

Yeah, I think it will be fine to remove the compatibility definition in project metadata.

https://github.com/weaveworks/grafanalib/blob/b312693fc2d5548e57c23a8bd9503ed8d15603fc/setup.py#L36

With kind regards,
Andreas.

amotl:
Hi again,

we submitted a corresponding patch to fix the project metadata.

- GH-652

With kind regards,
Andreas.

PR #652, opened 2024-03-16
chore: Update project metadata to reflect minimum Python version 3.8

amotl:
Dear maintainers,

coming from GH-650, this package is no longer compatible with Python 3.7 after starting to use `typing.Literal` recently. This patch accounts for that by adjusting the project metadata correspondingly.


JamesGibo (approved):
Thanks for updating the docs

PR #655, opened 2024-04-19
AlertRulev9 supports dashbord_uid and panel_uid.

JaSei:
## What does this do?
Add link from alert rule to the dashboard and panel.

## Why is it a good idea?
Those options was avaialable, but wasn't used and generated alert doesn't have connection to the dashboard and panel. By this fix it's possible.

### DIFF
--- a/grafanalib/core.py
+++ b/grafanalib/core.py
@@ -1696,6 +1696,12 @@ class AlertRulev9(object):
             else:
                 data += [trigger.to_json_data()]
 
+
+        if self.dashboard_uid:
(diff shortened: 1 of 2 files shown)


JamesGibo (changes requested):
Look good, not tested but does not look to be a breaking change.
Would you be able to add a test for this new annotation logic?

grafanalib/core.py
```diff
@@ -1719,6 +1719,12 @@ def to_json_data(self):
             else:
                 data += [trigger.to_json_data()]
 
+
```
JamesGibo, line 1722:
```suggestion
```

PR #659, opened 2024-05-10
bytemonk: Added tooltip sort to PieChartv2 panel

bytemonk:
## What does this do?
Added tooltipSort parameter to PieChartv2 panel

## Why is it a good idea?
This option was missing for the tooltips in PieChartv2 panel

## Context
Added tooltipSort parameter to PieChartv2 panel

## Questions
N/A


JamesGibo (approved):
Thanks for this great addition

PR #661, opened 2024-06-18
feat: add account to CloudwatchMetricsTarget

MattiasMTS:
## What does this do?
Adding the Account parameter to Cloudwatch datasource.

## Why is it a good idea?
A CloudWatch monitoring account views data from source accounts so you can centralize monitoring and troubleshooting activities across multiple accounts.


JamesGibo (approved):
LGTM - Thanks for your addition
I have updated the doc string which had a typo

PR #664, opened 2024-07-12
add mappings to field config for Table

kumikoda:
## What does this do?
This change enables you to configure value mappings for Table. Previously, the `mappings` argument provided would just get ignored.

## Questions
I am not sure about backwards compatibility, but I am using Grafana v9.4.7 and it expects `mappings` in this location.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ Changelog
 =========
 
 x.x.x ?
+* Fix mappings for Table
+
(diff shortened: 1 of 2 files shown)


JamesGibo (approved):
LGTM, thanks for your contribution

abalfanzatomic:
this also leaves in the OLD incorrect/duplicated mappings entry in Table .. @JamesGibo

PR #677, opened 2024-10-08
add alert list label filter option

kumikoda:
## What does this do?
Enable AlertList tables to filter by label


dswansonvizio (commented):
Just one note, I think you may also want to add a comment on the class itself with the new parameter.

PR #685, opened 2025-01-09
update timeseries to allow hiding axis grid

kumikoda:
## What does this do?
Allows hiding of axis grid in time series

## Why is it a good idea?
You can do it in the UI but not with grafanalib


casibbald:
Please kindly provide before and after screenshots

